PR #3, opened 2019-03-22
gradle settings

dominikl:
Adjust some gradle settings:
- Add a dependency for `javax.activation`. This is needed in order to be able to build with Java > 9
- Replace `omero-blitz` dependency with `omero-java-gateway`

--exclude


dominikl:
Split up into two PRs #4 and #5

PR #4, opened 2019-03-28
Add javax.activation dep

dominikl:
Add a dependency for `javax.activation`. This is needed in order to be able to build with Java > 9.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -32,7 +32,9 @@ dependencies {
     implementation("org.jfree:jfreechart:1.0.19")
     implementation("org.openmicroscopy:omero-blitz:5.5.0-m4")
     implementation("org.swinglabs:swingx:1.6.1")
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -32,6 +32,7 @@ dependencies {
     implementation("org.jfree:jfreechart:1.0.19")
     implementation("org.openmicroscopy:omero-blitz:5.5.0-m4")
     implementation("org.swinglabs:swingx:1.6.1")
+    implementation("javax.activation:activation:1.1.1")
```
jburel: maybe add If Java>=9

jburel:
This has already been tackled by @rgozim 
in the packaging work.

jburel:
see https://github.com/rgozim/omero-insight/blob/m5/build.gradle

jburel:
Re-opening this is better to have that change in asap since it blocks other work.

PR #5, opened 2019-03-28
Add omero-java-gateway dep

dominikl:
Replace `omero-blitz` dependency with `omero-java-gateway`.

--exclude  `omero-java-gateway.jar` not available on artifactory yet.


dominikl:
Gateway split not gonna happen yet.

PR #8, opened 2019-04-07
Version 

jburel:
Read the version from the manifest
The value not longer needs to be set from the configuration file
To test:
 * Run ``gradle build``
 * Unzip ``build/distributions/omero-insight-5.5.0-SNAPSHOT.zip``
 * go to the created directory
 * Run ``bin/omero-insight``
 * Check that the version is displayed on the splashscreen

When running ``gradle run``, no version should be displayed. In that case, it should be set to ``Unknown``

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -37,7 +37,6 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.ResourceBundle;
 import java.util.Set;
 import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
@@ -94,7 +93,6 @@ import ome.formats.importer.ImportLibrary;
 import ome.formats.importer.OMEROWrapper;
 import ome.formats.importer.util.ProportionalTimeEstimatorImpl;
 import ome.formats.importer.util.TimeEstimator;
-import ome.system.UpgradeCheck;
 import ome.util.checksum.ChecksumProvider;
 import ome.util.checksum.ChecksumProviderFactory;
 import ome.util.checksum.ChecksumProviderFactoryImpl;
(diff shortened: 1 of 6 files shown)


src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -1413,30 +1411,6 @@ ExperimenterData getUserDetails(SecurityContext ctx, String name,
...
 	}
 
-	/**
-	 * Returns <code>true</code> if an upgrade is required, <code>false</code>
-	 * otherwise.
-	 *
-	 * @param name The name of the agent.
-	 * @return See above.
-	 */
-	boolean isUpgradeRequired(String name)
```
mtbc: Ah, we can skip this because the Java Gateway handles it?

mtbc:
Note that the zip *used* to be called `OMERO.insight-*` rather than `omero-insight-*`. We may have to adjust various documentation.

Also I *do* need to be just above the `bin/` directory for the `omero-insight` script to work. The 5.4 `OMEROinsight_unix.sh` works from wherever I am.

mtbc:
Splash screen displays `5.5.0-SNAPSHOT` as expected. Though, if I try to log in to a current server then it still tells me that I need to download the latest 5.4. OMERO.insight.

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
jburel: yes

mtbc:
Yes, version is "Unknown" with `gradle run`.

mtbc:
None of my comments necessarily block this PR: they may all be expected or handled by other cards.

jburel:
For the renaming: I will create an issue and same for the script to launch

mtbc:
:+1: Everything else looks fine to me.

jburel:
For ``download the latest 5.4. OMERO.insight``,  something else must be kicking around outside insight code. I am planning to version the properties file this week

jburel:
The version https://github.com/ome/omero-blitz/pull/31 should have fixed the issue you reported. It looks like we have another problem somewhere. 
This is not related to this PR

jburel:
@mtbc found the reason it is b/c the version in ``build.gradle`` is still at m4 i.e. ``implementation("org.openmicroscopy:omero-blitz:5.5.0-m4")`` 
 I can bump that dependencies on that PR

mtbc:
This *does* let me log in to eel merge and west (with padlock closed).

jburel:
Great

PR #21, opened 2019-04-26
plugin bump

jburel:
Bump to m2 so that the version of executable is correctly set.

To test run either ``gradle packageApplicationDmg`` or ``gradle packageApplicationExe``
check that the generated distribution has the version ``5.5.0-SNAPSHOT``
a warning/error indicating that the version is "not correct" will be printed in the console but the distribution will be generated
There is no issue if a value like ``5.5.0`` is used

This PR also uses https://github.com/ome/omero-java-gateway instead of omero-blitz.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ dependencies {
     implementation("org.jfree:jfreechart:1.0.19")
     implementation("org.swinglabs:swingx:1.6.1")
 
-    implementation("org.openmicroscopy:omero-blitz:5.5.0-m5") {
+    implementation("org.openmicroscopy:omero-java-gateway:5.5.0-m2") {
         // Conflicts with `net.java.dev.jna`
         exclude group: "com.sun.jna", module: "jna"
         exclude group: "org.springframework"
--- a/build.gradle
+++ b/build.gradle
@@ -40,7 +40,7 @@ dependencies {
     implementation("org.jfree:jfreechart:1.0.19")
     implementation("org.swinglabs:swingx:1.6.1")
 
-    implementation("org.openmicroscopy:omero-java-gateway:5.5.0-m2") {
+    implementation("org.openmicroscopy:omero-java-gateway:5.5.0-m3") {
         // Conflicts with `net.java.dev.jna`
         exclude group: "com.sun.jna", module: "jna"
         exclude group: "org.springframework"
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -10,7 +10,7 @@ repositories {
 }
 
 dependencies {
-    implementation "org.openmicroscopy:omero-javapackager-plugin:5.5.0-m1"
+    implementation "org.openmicroscopy:omero-javapackager-plugin:5.5.0-m2"
 }
 
 gradlePlugin {


dominikl:
I ran both `packageApplicationDmg` and `packageApplicationDmg` (also `distZip`) on Debian 9. Can't find any dmg or exe, but I have a `omero-insight/build/distributions/OMERO.insight-5.5.0-SNAPSHOT.zip`. Extracted that and ran `bin/omero-insight`, but then get a:
```
...
Caused by: java.util.MissingResourceException: Can't find bundle for base name omero, locale en_GB
	at java.util.ResourceBundle.throwMissingResourceException(ResourceBundle.java:1581)
	at java.util.ResourceBundle.getBundleImpl(ResourceBundle.java:1396)
	at java.util.ResourceBundle.getBundle(ResourceBundle.java:782)
	at omero.gateway.Gateway.login(Gateway.java:1138)
```
But I can confirm, the `libs` directory contains the `omero-java-gateway-5.5.0-m2.jar`.

joshmoore:
Since https://github.com/ome/omero-java-gateway/pull/3/files#diff-626d029ff402dc027311598185b96d69L1138 wasn't merged?

jburel:
Now that https://github.com/ome/omero-java-gateway/pull/3 is sorted, we can review it tomorrow 
and tagged if merged
This should solve the issue mentioned above.

jburel:
This also leads to another point. Maybe the daily build should replace ``mX`` by ``SNAPSHOT`` so that the change can be tested

jburel:
@dominikl bump to m3. it should now work

joshmoore:
> Maybe the daily build should replace mX by SNAPSHOT so that the change can be tested

You mean that devspace should update insights dependencies?

dominikl:
👍  Works. Tested with `gradle distZip` because `packageApplicationDmg` (and `...Exe`) doesn't produce any OSX or Win builds, might be an issue of my system setup.

joshmoore:
`doesn't produce any OSX or Win builds` sounds worrisome.

dominikl:
Well, if I run these commands, I get a build directory with `classes  config  generated  install  libs  packaged  resources  scripts  tmp` sub directories with various stuff in it, but can't find a `*.dmg` or `*.exe` any where. Using debian9, openjdk8 and openjfx (for the the javapackager).

mtbc:
Maybe it builds packages suited only for the platform you're on?

jburel:
I don't think you can build the dmg on Debian9, you will need to test on your mac.
if on your mac you do not have the package manager it will not work
you will have to do a ``gradle clean`` before running the command to generate the dmg.
something does not seem correct with your setup

jburel:
> You mean that devspace should update insights dependencies?
yes similar to what happens for the omero-build submodules
The same will apply to omero-matlab.

pwalczysko:
After the build of insight is finished on Win7 laptop, I go into the build > packaged > main directory. There, I can see an insight icon. 
may2019

Note that my OS thinks this is a Firefox-executable html.

pwalczysko:
The executable is in build > packaged > main  > bundle directory. The version is 5.5.0 SNAPSHOT as expected.

pwalczysko:
The built insight is connecting to eel as expected. All good here.

PR #24, opened 2019-05-03
new method guava

jburel:
use method introduced in newer version of guava
cc @sbesson 

--exclude


jburel:
Closing this PR. This will be included in another PR with the "lighter" gateway

PR #26, opened 2019-05-08
Gateway

jburel:
This PR adjusts the code and add required dependencies so it can use the newer version of the java gateway
see https://github.com/ome/omero-java-gateway/pull/5

To test:
 * Download the jar from https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/
 * Launch insight.
 * Log in
 * Import an image


mtbc:
With https://github.com/ome/omero-java-gateway/pull/5 in the stack Insight seems to be working well. :+1: Only issue I see, probably unrelated to this PR, is an oddly fuzzy bird's eye view:
svs

jburel:
The fuzzy bird eyeview is due to the fact that we retrieve an existing thumbnail and "stretched" it. 
Not related

PR #27, opened 2019-05-08
version

jburel:
check version file


jburel:
https://github.com/snoopycrimecop/omero-insight/blob/south_merge_trigger/build.gradle has the correct value
Merging when travis is green

PR #28, opened 2019-05-10
Adjust sentence about unpacking the artifact.

mtbc:
The two sentences looked as if they needed connecting somehow.


jburel:
Thanks

PR #29, opened 2019-05-10
Have Insight default to connecting securely.

mtbc:
I propose that it's better to have the padlock start off locked. Try starting up Insight and see how it's now set.

### DIFF
--- a/src/config/container.xml
+++ b/src/config/container.xml
@@ -43,9 +43,9 @@
       <port>4063</port>
       <portSSL>4064</portSSL>
       <!-- Indicates the data transfer is encrypted (slower).
-      The default value is false
+      The default value is true
       -->
-      <encrypted configurable="true">false</encrypted>
+      <encrypted configurable="true">true</encrypted>
     </structuredEntry>
 
     <!-- the value corresponding to the agent defined as the master
--- a/src/config/containerImporter.xml
+++ b/src/config/containerImporter.xml
@@ -43,9 +43,9 @@
       <port>4063</port>
       <portSSL>4064</portSSL>
       <!-- Indicates the data transfer is encrypted (slower).
-      The default value is false
+      The default value is true
       -->
-      <encrypted configurable="true">false</encrypted>
+      <encrypted configurable="true">true</encrypted>
     </structuredEntry>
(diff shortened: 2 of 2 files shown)


jburel:
It certainly avoid the surprise we had in Singapore or with merge-ci

jburel:
cc @pwalczysko

mtbc:
Given caution raised by @jburel I experimented a little with how the `--encrypted` option affects CLI import of [77928.svs](https://downloads.openmicroscopy.org/images/SVS/77928.svs). The upload time slowdown for that was around 25%. I don't know how well one can generalize from that. (Heh, I notice I gave a HTTP**S** URL for the image file.)

pwalczysko:
> It certainly avoids the surprise we had in Singapore or with merge-ci

I would think, however unpleasant these problems were, they are rather team-internal. Not sure about the new behaviour - speed of import means a lot

jburel:
The speed of import is a major problem.
I think we should stick to the current default i.e. start with the padlock not locked

joshmoore:
I don't know how hard it would be to do this in OMEROGateway, but an alternative would be to move to SSL as the default for everything _except_ importing. This would work due to https://github.com/openmicroscopy/openmicroscopy/pull/5645 which is that if SSL is _not_ enabled, the importer natively falls back to TCP.

jburel:
I think we can look at that post 5.5 since it has been like that for a long time. So need to introduce extra work now

mtbc:
Best to close at this time then; if somebody wants to pick up the work then they can cherry-pick the commit.

Might be nice if paranoid people *can* choose to import over SSL too even if not default.

joshmoore:
> Might be nice if paranoid people can choose to import over SSL too even if not default.

That's the inverse logic, which would require checking. I _assume_ that setting the lock icon propagates down to import.

jburel:
the lock icon means that any data transfer is "secured" including import

joshmoore:
> the lock icon means that any data transfer is "secured" including import

Definitely know that's the intent. But with the automatic fallback, it probably would require a visual inspection of the log to know that it's working as intended.

PR #30, opened 2019-05-13
Imagej

jburel:
This fixes the distribution of the ImageJ plugin
To test
 * download the OMERO.imagej zip
 * unzip 
 * Place the folder in the Fiji plugin directory
 * Check that you have OMERO entries in the menu
 * Check that you can connect to OMERO.

cc @pwalczysko


pwalczysko:
Rebuilt the https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/21/ this evening with this PR.

Retest:

   - OMERO item is visible in Fiji
   -  connect successfully to eel
   - do the "crop workflow" from the training materials, full success, no problem to import the crop
   - do the import of a screenshot from local disk via Fiji Omero plugin, passes
   - draw a line on an image opened from OMERO via Fiji Omero plugin, then do line profile in Fiji, which is an image, import that image (=line profile), passes

It seems that the situation here improved very significantly.

:)

PR #33, opened 2019-05-21
Silently strip http(s):// from the server name

dominikl:
Many people accidentaly enter the full OMERO.web URL as server name in Insight. This PR simply strips the `^https?://` part.

**Test:**
Add a new Server. Use `https://demo.openmicroscopy.org` as server name. You should then be able to simply connect to `demo.openmicroscopy.org`.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/util/ui/login/ServerEditor.java
+++ b/src/main/java/org/openmicroscopy/shoola/util/ui/login/ServerEditor.java
@@ -33,6 +33,8 @@ import java.awt.GridBagLayout;
 import java.awt.Insets;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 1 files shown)


mtbc:
Maybe *first* trim it and put it all to lowercase (to catch HTTP...) and also trim any trailing "/"?

joshmoore:
This makes sense as an immediate ease-of-use improvement for the user, but there might be an exception here if we decide to make [omero-client.json](https://idr.openmicroscopy.org/connection/omero-client.json) standard API in the future.

cc: @manics

mtbc:
Could be in the longer term we just try throwing it at the `java.net.URL` constructor to see what sticks and take it as a bare host name if malformed.

dominikl:
Thanks Mark. Yes, using `java.net.URL` is a much better way to do this. Pushed another commit.

dominikl:
But I'm not sure what this  omero-client.json means... how that'd affect Insight.

src/main/java/org/openmicroscopy/shoola/util/ui/login/ServerEditor.java
```diff
@@ -587,11 +589,26 @@ String getSelectedServer()
 		if (row == -1) return null;
 		String v = (String) table.getValueAt(row, 1);
 		if (v == null) return null;
-		String trim = v.trim();
+		String trim = checkServerName(v);
```
mtbc: maybe pass `v.trim()` as the arg?

joshmoore:
> But I'm not sure what this omero-client.json means... how that'd affect Insight.

The values there could certainly also be beneficial for an insight connecting to that particular server (includes load-balancing, etc.)

pwalczysko:
One user in Cambridge also persistently entered a trailing slash. Can we cater for that too ?

mtbc:
If we do want to address the lone-trailing-slash issue maybe something like `while(s.endsWith("/")) s = s.substring..."`

mtbc:
(or just replace `"/+$"` with nothing)

PR #34, opened 2019-05-23
use release versions

jburel:
use release versions
Check that it is green


dominikl:
Looks good. Travis is green. Also built fine locally, all jars there with the correct versions 👍

PR #39, opened 2019-06-04
search workaround

jburel:
add all scope if none set
This change could be pushed down to https://github.com/ome/omero-gateway-java/blob/master/src/main/java/omero/gateway/facility/SearchFacility.java

cc @mtbc @dominikl


jburel:
closing the workaround now that we have identified the issue this will not be the most suitable approach

PR #40, opened 2019-06-04
handle null value

jburel:
handle null value
This was noticed on some images locally. The value returned was ``null``, this implies that the imported was constantly changing.


joshmoore:
@jburel : if it could change at any moment, you'd probably be better of capturing the results once and then reusing it:

```
Something i = i.getInserted();
if (i != null) {
    i.somethingElse();
}
```

jburel:
What is intriguing is why 
``
if (event == null || !event.isLoaded() || event.getTime() == null) return null
``
is true where event is ``getCreationEvent`` for an image

jburel:
@joshmoore happy with the last commit?

joshmoore:
WFM.

PR #44, opened 2019-06-11
feedback

jburel:
set url to send feedback

To test:
 * Send feedback using the feedback option
 * check that it is registered


mtbc:
With `gradle run` from my laptop I submitted https://www.openmicroscopy.org/qa2/qa/feedback/27619/.

jburel:
@mtbc thanks

PR #46, opened 2019-06-13
Record error

jburel:
Record error
Check that the application still runs 

cc @mtbc

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
@@ -473,6 +473,7 @@ public class DataServicesFactory
 	    
 		if (uc == null)
             throw new NullPointerException("No user credentials.");
+        LogMessage msg;
 		String name = (String) 
 		container.getRegistry().lookup(LookupNames.MASTER);
 		LoginCredentials cred = new LoginCredentials();
@@ -521,7 +522,8 @@ public class DataServicesFactory
             UpgradeCheck check = new UpgradeCheck(cs.getConfigValue("omero.upgrades.url"), clientVersion, name);
             check.run();
         } catch (ServerError e2) {
-            registry.getLogger().warn(this, "Could not access ConfigService");
+            msg = new LogMessage("Could not access ConfigService", e2);
+            registry.getLogger().warn(this, msg);
         }
 
         //Post an event to indicate that the user is connected.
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
```diff
@@ -473,14 +473,9 @@ public void connect(UserCredentials uc)
 	    
 		if (uc == null)
             throw new NullPointerException("No user credentials.");
+        LogMessage msg;
 		String name = (String) 
 		container.getRegistry().lookup(LookupNames.MASTER);
-		if (CommonsLangUtils.isBlank(name)) {
-            name = LookupNames.MASTER_INSIGHT;
```
mtbc: It's okay to move this step even though `cred.setApplicationName` won't then benefit from it?

jburel: Looking at the previous code, it was set as OMERO.insight and not insight that's why I moved it

jburel: sorry that line should be kept where it was

mtbc:
Is https://github.com/ome/omero-insight/pull/46/files#diff-d3089c1dff7a3f6c789980fcb8bcb74aL581 broken? I'm starting to wonder if safest would be to put the no-"OMERO." version into a different variable for the upgrade check rather than mutating `name` itself.

src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
```diff
@@ -490,8 +488,13 @@ public void connect(UserCredentials uc)
         cred.setCheckNetwork(true);
         cred.setCompression(determineCompression(uc.getSpeedLevel()));
         cred.setEncryption(uc.isEncrypted());
-        
-		ExperimenterData exp = omeroGateway.connect(cred);
+        ExperimenterData exp = null;
+        try {
+            exp = omeroGateway.connect(cred);
+        } catch(Exception e) {
+            e.printStackTrace();
```
mtbc: Is this intended?

jburel: no, I was trying to figure out why I could not connect to merge-ci

PR #47, opened 2019-06-14
Deprecate SERVER_5_4_8_OR_LATER.

mtbc:
Removes from Insight the special thumbnail handling for OMERO.server <5.4.8.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/LookupNames.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/LookupNames.java
@@ -115,8 +115,8 @@ public class LookupNames
 
     /** 
      * Field to check if the server version is 5.4.8 or later.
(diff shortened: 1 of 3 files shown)


dominikl:
I'm not really sure about this. Isn't `5.5.0` supposed to be backwards compatible with `5.4.x` ?

mtbc:
Oh, is it? Sorry, I didn't realize, in which case, yes, best to close this.

dominikl:
I'm not sure... just wondering because we now have the all the 'skip versioncheck' flags. @jburel ?

jburel:
you won't be able to connect to 5.4.x server with the current 5.5.x insight.

dominikl:
That's true... Gateway has the option to ignore version mismatch, but Insight actually doesn't. Guess that's alright then to remove this code.

jburel:
Thanks

PR #51, opened 2019-06-25
Help

jburel:
Update version in the help

The value should match the value displayed on the login screen

cc @mtbc

### DIFF
--- a/src/config/about.xml
+++ b/src/config/about.xml
@@ -7,11 +7,10 @@ and annotating microscope images and metadata.
 </p>
 <p>
 Version: @VERSION@<br></br>
-Build: @DATE@<br></br>
 </p>
 <p>
 <br></br>
-(c) Copyright 2006-2017 University of Dundee &amp; Open Microscopy Environment.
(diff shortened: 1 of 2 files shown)


mtbc:
This fails with `gradle run` but maybe that's expected? There's probably some other gradle thing I'm meant to use? I tried `gradle distTar` but can't find what it made, apparently nothing.

jburel:
What do you mean by fail? Did it show "unknow"? if it is the case, that is expected

mtbc:
It did.

mtbc:
I was hoping that I could do some other kind of build and get an actual number there? But maybe not?

jburel:
I did a ``gradle build`` and in that case the version does not seem to be picked. I will need to review it

jburel:
To test 
 * run ``gradle build``
 * unzip the ``build/distributions/OMERO.insight-5.5.3-SNAPSHOT.zip``
 * run ``build/distributions/OMERO.insight-5.5.3-SNAPSHOT/bin/omero-insight``
 * log in and go to help

PR #54, opened 2019-07-03
Bump gateway to 5.5.2

joshmoore:
In concert with the upcoming OMERO.server 5.5.1

PR #56, opened 2019-07-05
Bump to 5.5.3

joshmoore:
Follows on from gh-54 by bumping GW _further_. Version bump for omero-insight and changelog are also included.


mtbc:
@sbesson: From Fedora 30 Insight 5.5.2 cannot connect to demo but current 5.5.3-SNAPSHOT can.

Issue #57, opened 2019-07-08
Hangs when importing very many files: servants held open


mtbc:
[[image.sc] OMERO.importer 5.4.10 throws Ice.ObjectNotExistException when importing hundreds of files](https://forum.image.sc/t/27124) describes investigation into "servantsPerSession reached" and UI freeze when importing over 3,000 files. Probably need to see why upload is holding servants open.

jburel:
Closing issue fixed

PR #58, opened 2019-07-10
Import ui fixes

dominikl:
Fixes the importer UI freeze when importing many images and enable the cancel, retry, submit buttons again. Together with https://github.com/ome/omero-gateway-java/pull/16 this should fix the issue https://github.com/ome/omero-insight/issues/57 .

**TEST**
Import > 4000 images on Windows. The issue affected all OS, but on Windows it was especially bad.
Should also be tested with just a couple of images, including multi-file images (eg. *.dv + *.log) to check if the 'normal' importer UI still works as expected.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/ImagesImporter.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/ImagesImporter.java
@@ -136,7 +136,6 @@ public class ImagesImporter
         msg.print(exc);
         registry.getLogger().error(this, msg);
         registry.getUserNotifier().notifyError("Data Import Failure", s, exc);
-		viewer.cancelImport(loaderID);
 	}
 
     /**
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/TagsLoader.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/TagsLoader.java
@@ -113,7 +113,6 @@ public class TagsLoader
         registry.getLogger().error(this, msg);
         registry.getUserNotifier().notifyError("Data Retrieval Failure", 
                                                s, exc);
-		viewer.cancel();
 	}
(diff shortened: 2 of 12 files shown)


dominikl:
It works, but I think I still need the `SwingUtilities.invokeLater` to handle that properly.

rgozim:
On Windows, I was able to import ~400 real images into user-4 on merge-ci (I don't have a sample size of 4000 images). The client machine was under full load for a number of hours and didn't have any issues. Should I test this with 4000 fake images?

dominikl:
Without the bugfix you would have seen a signifcant slow down with only 400 images too, but it might still have worked. Would be good to test it with 4000 images too (that definitely didn't work). I guess fake images would do. I usually generate a large number of small pngs with a small bash script using Imagemagick:
```
#/bin/bash

for n in `seq $1`; do
        OUT=`printf 'img-%05d.png' $n`
        convert -size 100x100 -background blue -fill green \
                -gravity center -pointsize 150 \
                label:$n $OUT
done
```

mtbc:
Note you also have some little PNGs in https://github.com/ome/omero-common/tree/v5.5.2/src/test/resources that you can just `ln` to a bunch of other names.

rgozim:
Managed to import around ~3900 images before merge-ci shutdown with no errors observed (on a mac). I'll try again later on a Windows machine.

joshmoore:
> Managed to import around ~3900 images

:+1: Can you also re-try each of the same tests without this PR and see if you see the hangs? Cheers.

jburel:
Tested import with and without this PR
set-up:
* folder with 4000 png
* MacOSX

Test 1 (without this PR)
* The UI starts to freeze around 2000 files


Test 2 (with this PR)
* The import goes smoothly.

Thanks @dominikl

jburel:
* Cancel All button is now available on the import tab.
Merging

PR #62, opened 2019-08-06
Import failure

jburel:
indicate failure when connection issue
To test: 
 * log in as a group owner
 * try to import for another user
 * the import will fail
 * Check that a message "failed to upload" is displayed and that you can view the exception
cc @pwalczysko


src/main/java/org/openmicroscopy/shoola/env/data/util/Status.java
```diff
@@ -367,6 +370,12 @@ public void markedAsDuplicate() {
         firePropertyChange(DUPLICATE_PROPERTY, null, DUPLICATE_PROPERTY);
     }
 
+    /** Marks that the upldoad failed due to connection issue. */
```
mtbc:
```suggestion
    /** Marks that the upload failed due to connection issue. */
```

jburel:
so the problem must be at the writing in the repository level

PR #53, opened 2019-06-26
Imagej script

jburel:
This PR adds the ability to use the session ID and the selected objects in the UI in script 
to test:
 * Copy the imagej plugin into the plugins directory of Fiji
 * Connect to OMERO
 * Select an image. If you select any other object, only the object type should be printed out after connection
 * In Fiji go to ``File>Script ...``
 * Select ``Language > Groovy``
 * Run the following code snippet
```
// OMERO Dependencies
import omero.gateway.Gateway
import omero.gateway.LoginCredentials
import omero.gateway.SecurityContext
import omero.gateway.facility.BrowseFacility
import omero.gateway.facility.DataManagerFacility
import omero.gateway.model.DatasetData
import omero.log.SimpleLogger

import ome.formats.importer.ImportConfig
import ome.formats.importer.OMEROWrapper
import ome.formats.importer.ImportLibrary
import ome.formats.importer.ImportCandidates
import ome.formats.importer.cli.ErrorHandler
import ome.formats.importer.cli.LoggingImportMonitor

import loci.formats.in.DefaultMetadataOptions
import loci.formats.in.MetadataLevel

import ij.IJ

// Setup
// =====

// OMERO Server details
HOST = org.openmicroscopy.shoola.MainIJPlugin.getHost()
PORT = org.openmicroscopy.shoola.MainIJPlugin.getPort()
group_id = "-1"
USERNAME = org.openmicroscopy.shoola.MainIJPlugin.getSessionId()
PASSWORD = org.openmicroscopy.shoola.MainIJPlugin.getSessionId()

def connect_to_omero() {
    "Connect to OMERO"

    credentials = new LoginCredentials()
    credentials.getServer().setHostname(HOST)
    credentials.getServer().setPort(PORT)
    credentials.getUser().setUsername(USERNAME.trim())
    credentials.getUser().setPassword(PASSWORD.trim())
    simpleLogger = new SimpleLogger()
    gateway = new Gateway(simpleLogger)
    gateway.connect(credentials)
    return gateway

}

def open_image_plus(HOST, USERNAME, PASSWORD, PORT, group_id, image_id) {
    "Open the image using the Bio-Formats Importer"

    StringBuilder options = new StringBuilder()
    options.append("location=[OMERO] open=[omero:server=")
    options.append(HOST)
    options.append("\nuser=")
    options.append(USERNAME.trim())
    options.append("\nport=")
    options.append(PORT)
    options.append("\npass=")
    options.append(PASSWORD.trim())
[...]

PR #66, opened 2019-08-21
Make Insight login work with websocket URLs

dominikl:
With this PR:
- A URL can be entered as hostname
- Port can be empty (default is set in the java gateway) (depends on https://github.com/ome/omero-gateway-java/pull/21 )

## Test
Connect to IDR next via websocket URL`wss://[OMERO SERVER]/omero-ws` (this again depends on https://github.com/ome/omero-blitz/pull/66 ) . Connect to another server using usual hostname and port.

## Note
There's actually one caveat: This PR slightly changes the appearance of the server list dialog: There are no separate fields any longer for hostname and port. If a port other than default port is used, one has to add it with a `:` after the hostname. Will probably need updating various help and training docs /cc @pwalczysko .
Screen Shot 2019-09-02 at 09 56 16


snoopycrimecop:
Conflicting PR. Removed from build [OMERO-insight-push#138](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-push/138/). See the [console output](https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-push/138/consoleText) for more details.
Possible conflicts:
  - PR #53 jburel 'Imagej script'
    - src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java

--conflicts

src/main/java/org/openmicroscopy/shoola/util/ui/login/ScreenLogin.java
```diff
@@ -546,7 +546,7 @@ private void initialize(String userName, String hostName)
 					serverName = i.next();
 					if (k == n) {
 						value = servers.get(serverName);
-						if (value != null) {
+						if (value != null && value.length() > 0) {
```
mtbc: `CommonsLangUtils.isNotBlank` can be handy for this kind of thing.

dominikl: Not sure, thought we want to gradually get rid off external apache/google dependencies?

mtbc: Where the JDK now offers an equivalent method, sure. For our heavyweight full-application JARs I wasn't aware of a broader push but I may have missed something? I'd have thought they were worth it for readability and reusing others' work but happy to go in the other direction if that's what's been decided. Certainly not a blocker for this PR anyway.

mtbc:
Initial functional testing looking promising. Can't see any data on idr-next but also can't via 4064 so maybe expected.

manics:
Works fine to me too, for idr had to play around with selecting/unselecting users/groups to view the data.

I couldn't find the insight logs (OS X, the menu option doesn't load anything). I wanted to check the error message was informative when using an incorrect URL, but that's unrelated to this PR.

dominikl:
That's a good point. The `omeroinsight.log` file is usually in `~/omero/log`. Opening the directory from the menu option in Insight works for me.

dominikl:
Sorry, there's actually way more to it. The server editor / dialog still has the two cells (one for hostname and one for the port); unbelievable complex to remove that. 
--exclude for now

mtbc:
I suppose it would be too optimistic to let the port number imply the protocol?

dominikl:
The problem is just the UI, this server list/editor is a monster (re complexity)  :-/

dominikl:
This PR got much bigger than I intended initially, now also:
- Rewrites the ServerEditor: Using simple JList instead of complicated table
- Removes the group stuff from ScreenLogin dialog ( https://github.com/ome/omero-insight/pull/66/files#diff-7eacc67f747c3a2db767bc67d29ba011L201 ). Couldn't find anything where this would be used.
- Removes the unneccassary LoginCredentials class (and the various conversions between this class and UserCredentials)

manics:
Configuration:
Screen Shot 2019-09-05 at 16 39 11
Quit, then reload Insight. The configuration is messed up:
Screen Shot 2019-09-05 at 16 41 40

manics:
@dominikl Can you remove the exclude commdent from https://github.com/ome/omero-insight/pull/66#issuecomment-527154000

dominikl:
Argh, sorry, yes forgot about the exclude in the comment above; removed now.

manics:
The new functionality works well. I've got a UI issue on `OS X 10.13.6` `openjdk version "1.8.0_212"`: When `+` is clicked the `Add new server` dialogue is behind the `Servers` dialog:

Screen Shot 2019-09-09 at 11 50 34

jburel:
This certainly allows to have the same server name with different port.
This is major change in the UI configuration.
Doc needs to be updated. The message could be added to "new server dialog"
cc @pwalczysko

pwalczysko:
Yes, but the addition of the port manually is adding a burden on each and every insight user - no user now will be able to ignore the port, it will never be prepopulated, do I get it right ? Thus, the user will have to know it is "colon 4064" 
I am not sure that this is a wise move

dominikl:
No, if it's the default port (which probably is the case for the vast majority of the users) you don't have to enter a port, ie `demo.openmicroscopy.org` will do. Oh, but that reminds me about the case people entering `https://demo.openmicroscopy.org`, that's probably gonna mess things up now.

src/main/java/org/openmicroscopy/shoola/util/ui/login/ServerEditor.java
```diff
@@ -214,12 +204,20 @@ public void actionPerformed(ActionEvent e)
 		removeButton.setEnabled(!this.servers.isEmpty());
 		addButton.addActionListener(new ActionListener() {
 			public void actionPerformed(ActionEvent e) {
-				String add = JOptionPane.showInputDialog(ServerEditor.this, "", "Add new server",
+				String add = JOptionPane.showInputDialog(ServerEditor.this,
+						"Enter hostname or IP address (:port as needed)\n(e.g. test.openmicroscopy.org or 192.168.0.1:1234",
```
jburel: the port example should be indicated when the hostname is changed. It will probably be missed when the IP is used

pwalczysko:
Screen Shot 2019-09-23 at 12 52 28


See above - some servers have ports listed, some not. This is probably because I had the list of servers "from old insight" and all these have ports explicitly listed. But, when I entered a new server (nightshade) I have not listed the port with a colon. But, the port number was not automatically added either. This could be explained in the dialog above, where it just says "Enter a new server or use existing one" - not very informatinve actually.

pwalczysko:
Screen Shot 2019-09-23 at 12 55 39

This should actually say "Connecting to" - at the moment when I am dealing with the dialog here, I am not connected yet.

pwalczysko:
Discussed with @dominikl - the solution might be to put the info in the first dialog, 

Screen Shot 2019-09-23 at 13 35 36


There is space for it and maybe it can be formatted as well.
I would imagine 3 separate lines with 3 examples, introduced by 
``Server address can be entered as:``
1. my.server.name (default case)
2. my.server.name:port (for non-default ports)
3. 124.133.0.343:1245 (if you have an IP address)
4. websocket-example  (connecting via websocket)

pwalczysko:
Screen Shot 2019-09-25 at 11 11 35


That is much better, couple of further suggestions:

- remove the e.g. , maybe replacing it by making the header to read ``Server address examples``
- swap the italicising of the brackets with the addresses (i.e. italicise the address examples, leave the explanations non-italics) -> see how that looks like please though
- make either quite a few spaces between the server address and the explanation, or, even better, make it so that the explanations are indented in the same way for all 4 examples

dominikl:
Had to fix some more issues, hence some more tests needs to be taken:

1) Apart from testing with an already existing list of servers, also delete (or temporary rename) your existing Insight preferences (`rm -rf ~/.java/.userPrefs/org/openmicroscopy`), then test again. Make sure you're connecting by default to 'localhost' (and there is exactly one entry in the serverlist 'localhost').

2) Delete preferences again. Open the container.xml and enter a different default server.
Check that this (and only this) server shows up as default server on the login dialog.

3) Delete preferences again. Open the container.xml  and set the configurable flag to false. Check that the specified server shows up as default, and that the 'Server Editor' button is disabled.

src/main/java/org/openmicroscopy/shoola/util/ui/login/ScreenLoginDialog.java
```diff
@@ -24,6 +24,8 @@
 
 
 //Java imports
```
mtbc: maybe just get rid of these comments

pwalczysko:
All works as expected here. Also tested import using websocket into the pub-omero server, went fine.

Ready to merge FMPOV

jburel:
Breaking UI change so this needs to go in for 5.6

jburel:
@pwalczysko The import guide will need to be adjusted to indicate how to change the port.

jburel:
https://github.com/ome/omero-guide-upload/issues/1

PR #68, opened 2019-08-21
Added a fat jar distribution of imagej for fiji

rgozim:
Adds a fat jar distribution for FIJI

- run `gradle installFijiDist` to build the Fiji distribution
- test with fiji

PR #71, opened 2019-08-26
exclude jna

jburel:
exclude. This creates error in Fiji

To test on Window:
 * use the plugin in a freshly downloaded fiji .
 * test using imagej
 * test using the standalone build
 * Check that the open with option still works for the general build

cc @pwalczysko


pwalczysko:
It fails for me with the same error as before on Win7 Fiji freshly downloaded.

pwalczysko:
I also tried a legacy (2017) Fiji with your fix. Also the same legacy Fiji with 5.5.3 release inisght-ij. Both failed with the same error.

jburel:
This change by itself is not enough.
This needs to be done in conjunction with #68

PR #74, opened 2019-08-28
Update link to BF

jburel:
bump link to latest BF version

cc @pwalczysko 

To check the link
* Install the insight-imageJ plugin into the plugins directory of ImageJ
* Check that you do not have bio-formats jar or loci jar in the plugins directory
* Try to connect to OMERO
* A dialog will pop up indicating that the dependencies are not installed. 
* Check the link

### DIFF
--- a/src/config/container.xml
+++ b/src/config/container.xml
@@ -363,7 +363,7 @@
       <dependencies>bioformats_package.jar, loci_tools.jar</dependencies>
       <conjunction>or</conjunction>
       <directory>plugins</directory>
-      <info>https://downloads.openmicroscopy.org/latest/bio-formats5</info>
+      <info>https://downloads.openmicroscopy.org/latest/bio-formats6</info>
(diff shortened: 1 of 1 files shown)


jburel:
One thing we do not check if the version of the jar even if it is present. This should probably be checked (not in this PR)

pwalczysko:
Principally it works. But, the link points to https://downloads.openmicroscopy.org/bio-formats/6.2.1/ - then you need to click on artifacts, then - see screenshot - not very user firendly, hard to find what to download, the object is not on the screen, need to scroll and find it in the middle of the output.
Screen Shot 2019-09-03 at 10 43 29

Could we maybe rather point to B-F downloads on ope...py.org ?

jburel:
@pwalczysko could you have a look?

pwalczysko:
Tested with the build from merge-ci

Works as expected
Ready to merge

PR #75, opened 2019-08-30
Add config to jar

jburel:
Second phase
This PR includes config folder in the fat jar. This should make it easier for users to install the plugin
if you have previously installed omero.imagej in the plugins directory of ImageJ/Fiji, you first need to delete it.
To test with Fiji and ImageJ
 * If you build locally, you will under ``build/libs``, the ``omero_ij*.jar``. 
 * If you retrieve it from merge-ci, unzip ``imagej*.zip`` and retrieve the jar.  (When we tag, the ``zip`` will not be pushed, the ``jar`` will.
 * Place the omero_ij jar into the plugins directory.
 * Check that you can connect and view images etc.
 * Disconnect
 * ImageJ only, Check that there is no file ``logFileName_IS_UNDEFINED`` created in the folder containing ImageJ

cc @pwalczysko


dominikl:
Looks good. Tested with latest Fiji on Windows 10. Only copied the jar into the plugins folder.
Got a `[WARN] UPGRADE AVAILABLE:Please upgrade to 5.5.1.` but apart from that worked fine.

dominikl:
The config folder is still part of the zip, too. That might be a bit confusing for the users.

dominikl:
ImageJ is ok too, didn't see any `logFileName_IS_UNDEFINED` file.

jburel:
@dominikl the zip should no longer be uploaded. 
I will remove the config from the zip in case some uses it

jburel:
@dominikl done

dominikl:
Tested with a local build because of the travis error. No config directory included any longer, and plugin still works as expected. Good to merge from my point of view 👍

jburel:
Travis is green again.
Merging so that @dominikl can fix the conflicts in his PR

PR #88, opened 2019-10-24
Fix NPE

dominikl:
Return an empty list instead of null for `BrowserComponent.getSelectedDataObjects()`.
That should fix https://www.openmicroscopy.org/qa2/qa/feedback/27973/ and prevent similar potential NPEs.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/treeviewer/view/TreeViewerComponent.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/treeviewer/view/TreeViewerComponent.java
@@ -603,7 +603,7 @@ class TreeViewerComponent
 				setLeaves((TreeImageSet) display, s);
 				db = DataBrowserFactory.getDataBrowser(display);
 				list = browser.getSelectedDataObjects();
-				if (list != null && list.size() == 1) {
+				if (list.size() == 1) {
 					app = TreeViewerFactory.getApplications(
 					model.getObjectMimeType(list.get(0)));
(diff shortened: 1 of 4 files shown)


dominikl:
Hm, could have actually removed all the checks for null then too...

src/main/java/org/openmicroscopy/shoola/agents/treeviewer/view/TreeViewerComponent.java
```diff
@@ -3632,7 +3632,7 @@ public void download(File folder, boolean override, List<DataObject> selection)
             return;
         List l = selection == null ? browser.getSelectedDataObjects()
                 : selection;
-	    if (l == null) return;
+	    if (l.isEmpty()) return;
```
mtbc: This could be `null`.

mtbc: Can always use `Collections.emptyList()` as a handy constant.

dominikl: Really, how? If the `selection == null` then `browser.getSelectedDataObjects()` which is now an empty list instead of `null`.

mtbc: Oh doh, must have misread it first time, you're completely right - sorry for the noise!

Issue #86, opened 2019-10-24
"Export as ome-tiff" option should be disabled for large images


dominikl:
Large image can't be exported as ome-tiff. The button shoudl be disabled.
See https://www.openmicroscopy.org/qa2/qa/feedback/27969/

dominikl:
Can't reproduce, button is disable for large images.

dominikl:
If I remember correctly the criteria is big image == pyramidal image. If the server reports that the image doesn't need pyramids, but then refuses to export as ome-tiff, then there's not really much to do on the client side; expect hard coding some fixed value to disable the button.
But I can try again to reproduce with a non-default setting just for the record.

mtbc:
I guess, the config service allows the client to ask their values?

PR #89, opened 2019-10-30
Disable ome-tiff export for 'large' images

dominikl:
See title. Fixes https://github.com/ome/omero-insight/issues/86 , resp. https://www.openmicroscopy.org/qa2/qa/feedback/27969/

To reproduce: Increase the max plane size on the server (`omero.pixeldata.max_plane_height` , `omero.pixeldata.max_plane_width`). Without the PR the ome-tiff export button becomes enabled, because it's not considered as big image. 'big' image is `omero.pixeldata.max_plane...` in contrast to 'large' image which is  `omero.client.download_as.max_size`. With the PR `omero.client.download_as.max_size` will be taken to decide if the button should be enabled (like it already was the case for png, jpeg etc.).

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/metadata/editor/ToolBar.java
@@ -187,6 +187,10 @@ class ToolBar
         downloadItem.setEnabled(b);
         saveAsMenu.add(downloadItem);
 
+		Object ho = model.getRefObject();
+		boolean enabled = (ho instanceof ImageData ||
+				ho instanceof WellSampleData || ho instanceof DatasetData) && model.isExportable();
+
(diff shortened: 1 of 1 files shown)


pwalczysko:
Note: the image from the QA report is really very big:

208896 x 98000 -  this makes some 20471808000 pixels. Nothing even close to it is allowed by insight at present to be exported as ome-tiff. They must have changed the default values on the server imho. https://www.openmicroscopy.org/qa2/qa/feedback/27969/

will-moore:
The ```omero.client.download_as.max_size``` value is meant to reflect the maximum size that the rendering engine can handle, since we render a plane when we download_as PNG, TIFF or JPEG.
I think the default value is ~ 12k x 12k.
I'm not sure it should be used for OME-TIFF export but I don't know what the limiting factor is there.

PR #97, opened 2019-12-12
Fix version error message

dominikl:
If you connect with current Insight 5.5.6 to for example merge-ci you get an error message like this:

Screen Shot 2019-12-12 at 11 36 29

It says Insight 5.5.4 because it gets the version of the java-gateway. This PR simply fixes the error message, so that it correctly states the version of Insight not the gateway one. Also removes the "Please download..." sentence, as we can't automatically determine any more which Insight version will be compatible with which server version.

See also https://forum.image.sc/t/omero-python-3-5-6-milestone-for-upgrade-testing/32120/5

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
@@ -328,12 +328,12 @@ public class LoginServiceImpl
             text = "Error: System Failure.";
             break;
         case LoginService.VERSION_MISMATCH:
-            String cv = "";
-            try {
-                cv = failureDetails.split("\\s")[2];
-            } catch (Exception e) {
-                // just ignore
-            }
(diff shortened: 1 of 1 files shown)


jburel:
DomPR

src/main/java/org/openmicroscopy/shoola/env/data/login/LoginServiceImpl.java
```diff
@@ -328,12 +328,12 @@ public void notifyLoginFailure()
...
             break;
         case LoginService.VERSION_MISMATCH:
-            String cv = "";
-            try {
-                cv = failureDetails.split("\\s")[2];
-            } catch (Exception e) {
-                // just ignore
-            }
+            String cv = this.getClass().getPackage().getImplementationVersion();
+            if (cv == null)
```
jburel: handle empty string?

jburel:
Thanks @dominikl

PR #101, opened 2019-12-18
Fix two minor issues in ServerEditor

dominikl:
Just fixes two small issues with the server editor @pwalczysko noticed while adjusting the docs https://github.com/ome/omero-guide-upload/pull/3 .
- Now it's only possible to select a single entry in the server list (multi selection doesn't really make sense here).
- And you can select and delete the first server in the list.

Not a blocker for the current Insight release.


jburel:
I tried multiple key combination and only one entry can be selected
Merging this PR so it can be included in the coming release

Issue #55, opened 2019-07-04
Failed to retrieve icon


dominikl:
If you startup Insight a couple of ERROR messages appear in log file, although they don't seem to have any visible implication in the UI. (OSX, Insight 5.5.2)
```
2019-07-04 13:19:20,172 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/red_dot8.png.
2019-07-04 13:19:20,173 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/nuvola_knotes8.png.
2019-07-04 13:19:20,200 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/knimeIcon16.png.
2019-07-04 13:19:20,200 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/imageJ16.png.
2019-07-04 13:19:20,200 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/viewer16.png.
2019-07-04 13:19:20,201 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/eclipse_copy_edit16.png.
2019-07-04 13:19:20,202 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/eclipse_paste_edit16.png.
2019-07-04 13:19:20,202 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/eclipse_delete_edit16.png.
2019-07-04 13:19:20,202 ERROR [    o.o.s.agents.dataBrowser.IconManager] (entQueue-0) Failed to retrieve icon: <classpath>/images/agents/dataBrowser/nuvola_editcut16.png.
[...]

PR #104, opened 2020-01-21
fix missing icons

jburel:
Folder was renamed to ``databrowser`` see https://github.com/ome/omero-insight/tree/47e40da08b9cff563df9894bc5490c15b545b17a/src/main/resources/images/agents/databrowser
but the value was not updated in the configuration file.
See https://github.com/ome/omero-insight/issues/55

cc @dominikl


dominikl:
👍 Makes sense, thanks @jburel . The warning messages are gone with this PR.

PR #106, opened 2020-01-22
IJ plugin open image when using websocket

jburel:
Fix issue preventing to open images using B-F when using the imagej plugin

To test:
 * Download the ``omero_ij`` jar
 * connect to a server with ws enable e.g. workshop or outreach
 * Open an image in ImageJ
 * The image should open 
cc @pwalczysko @dominikl


dominikl:
Thanks @jburel . I thought I changed the `getHostname()` call everywhere back then, but missed that one apparently. I'll give it a try later.

dominikl:
Could reproduce the error with released version 5.5.8. Works with 5.5.9-Snapshot from merge-ci. Good to merge 👍

PR #113, opened 2020-03-12
Provide error message if unsupported protocol was used

dominikl:
See Gateway PR https://github.com/ome/omero-gateway-java/pull/34 , respectively issue https://github.com/ome/omero-gateway-java/issues/33 .

**Test**
Check that 'normal' login (hostname) and websocket login still works.
Check that you get a sensible error message if you try something like 'smb://someserver.org' .

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/ui/login/ServerEditor.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/ui/login/ServerEditor.java
@@ -414,15 +414,32 @@ public class ServerEditor
 	 */
 	List<String> getServers()
 	{
-    	Preferences prefs = Preferences.userNodeForPackage(ServerEditor.class);
-        String servers = prefs.get(OMERO_SERVER, null);
-        if (servers == null || servers.length() == 0)
-            return new ArrayList<>();
-        String[] l = servers.split(SERVER_NAME_SEPARATOR, 0);
+		Preferences prefs = Preferences.userNodeForPackage(ServerEditor.class);
+        List<String> servers = getServersFromPrefs(prefs);
+
+        // Workaround to load old preferences (remove again in next release):
+		Preferences oldPrefs = Preferences.userNodeForPackage(org.openmicroscopy.shoola.util.ui.login.ServerEditor.class);
+		servers.addAll(getServersFromPrefs(oldPrefs));
+
+		return servers;
+	}
+
+	/**
+	 * Returns the list of existing servers from the provided Preferences
+	 *
+	 * @param prefs The Preferences
+	 * @return See above.
+	 */
+	private List<String> getServersFromPrefs(Preferences prefs)
+	{
+		String servers = prefs.get(OMERO_SERVER, null);
+		if (servers == null || servers.length() == 0)
+			return new ArrayList<>();
+		String[] l = servers.split(SERVER_NAME_SEPARATOR, 0);
 		for (int i=0; i<l.length; i++)
 			l[i] = l[i].replace(":"+DEFAULT_PORT, "");
-        if (l == null)
-        	return new ArrayList<>();
+		if (l == null)
+			return new ArrayList<>();
 		return Stream.of(l).collect(Collectors.toList());
 	}
 	
--- a/src/main/java/org/openmicroscopy/shoola/env/ui/login/ServerEditor.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/ui/login/ServerEditor.java
@@ -419,7 +419,10 @@ public class ServerEditor
 
         // Workaround to load old preferences (remove again in next release):
 		Preferences oldPrefs = Preferences.userNodeForPackage(org.openmicroscopy.shoola.util.ui.login.ServerEditor.class);
-		servers.addAll(getServersFromPrefs(oldPrefs));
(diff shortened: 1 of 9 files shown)


dominikl:
Actually, in Insight I should not just pass on the error message, but use something more helpful for the user. I'll change that.

src/main/java/org/openmicroscopy/shoola/util/ui/login/ScreenLogin.java
```diff
@@ -62,6 +62,7 @@
 import javax.swing.event.DocumentListener;
 
 import org.openmicroscopy.shoola.env.config.OMEROInfo;
+import org.openmicroscopy.shoola.env.config.Registry;
```
jburel: something is not quite right in this class.
``env.*`` classes should not have leaked in this more general package.

dominikl: Looks like I introduced these imports when overhauling the login screen / server editor. Hm, gonna need quite some workarounds to get rid off them.

dominikl:
Just moved the `login` package from `util` to `env` as there are already other login / start up related ui classes like SplashScreen in `env` too.

jburel:
Screenshot 2020-04-02 at 19 40 53

I am not convinced a user will under the message " no enum ..."
It will be better to have a message like ``protocol not supported: FTP``

dominikl:
It looks like it wasn't build with my gateway PR https://github.com/ome/omero-gateway-java/pull/34 then. With that PR it will then say "XYZ is not supported". Thought the merge build contains all open PRs of all repositories?

jburel:
I will try again

jburel:
The dialog pops up twice with the same message

jburel:
Screenshot 2020-04-03 at 08 39 50
I was using the branch directly while trying to debug a connection issue. The message is better now

mtbc:
error

Works well for me too but could have been more timely when I entered the server name rather than after I'd added it to the list and actually tried to connect with my username and password. Still an improvement though!

dominikl:
Moving of the login package to util has probably caused that one looses all the previous server settings because the java prefs location is determinded by the classname. I'll add another commit to fix that.

jburel:
@mtbc when the server name is entered, there is no interaction with the gateway at that stage
Changing that will add complexity not really needed at this stage

dominikl:
The problem with the previous server list being lost is indeed due to moving the ServerEditor class into a different package. Shall I just close the PR? Not worth spending lots of time on this.

jburel:
Let's fix the protocol error notification problem in this PR
We can look at re-organising the package in another PR. 
I should have noticed that before

dominikl:
Added a simple workaround to load the old server list. This can simply be removed again in a later release.

src/main/java/org/openmicroscopy/shoola/env/ui/login/ServerEditor.java
```diff
@@ -414,15 +414,32 @@ void addRow(String server) {
...
 	{
-    	Preferences prefs = Preferences.userNodeForPackage(ServerEditor.class);
-        String servers = prefs.get(OMERO_SERVER, null);
-        if (servers == null || servers.length() == 0)
-            return new ArrayList<>();
-        String[] l = servers.split(SERVER_NAME_SEPARATOR, 0);
+		Preferences prefs = Preferences.userNodeForPackage(ServerEditor.class);
+        List<String> servers = getServersFromPrefs(prefs);
+
+        // Workaround to load old preferences (remove again in next release):
```
joshmoore, line 420: That's only true if the user definitely installs this version, correct?

dominikl: That's true... maybe 'next release' is a bit too optimistic.

dominikl: It's a bit pity that one needs the class to really exist to load the Preferences, afaik not possible to just load by name.

jburel:
Screenshot 2020-04-08 at 11 49 08
The list is back

jburel:
I will include that PR  and prepare a release of gateway next week. Then bump the version

jburel:
Small problem I noticed

The latest server is not remember e.g. 
* I connect to workshop
* disconnect
* start insight again
* localhost is displayed

dominikl:
Sorry, totally missed the last comment. I'll have a look at the issue today.

dominikl:
I think the issue https://github.com/ome/omero-insight/pull/113#issuecomment-613278827 was introduced with #66 . Trying to find out how, then I can fix it in this PR.

jburel:
@dominikl do you want to fix in this PR?
I think I will do a release as soon as https://github.com/ome/omero-insight/pull/137 is in

dominikl:
https://github.com/ome/omero-insight/pull/113#issuecomment-613278827 should be hopefully fixed now by the last commit.

jburel:
The latest server is selected. The last commit works.
I still have cases when the dialog pops again after clicking "ok"
The first time it works fine, the second time it shows up twice.

dominikl:
There was some weird code which had the effect the ActionListener was added multiple times. So this should be fixed now too.

jburel:
Looks good now

PR #114, opened 2020-03-12
Fiji: Crop option

jburel:
select the crop option if large image
to test: 
* Open a small image in Fiji, the crop option should not be selected
* Open a large image (image requiring pyramid), the crop option should be selected

cc @dominikl

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/ui/TaskBarManager.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/ui/TaskBarManager.java
@@ -361,8 +361,8 @@ public class TaskBarManager
 			buffer.append(ctx.getGroupID());
 			buffer.append("\niid=");
 			buffer.append(id);
-			buffer.append("]");
-			buffer.append("view=Hyperstack "); // select hyperstack by default
+			buffer.append("] ");
+			buffer.append("view=Hyperstack windowless=false "); // select hyperstack by default
 			if (largePlane) {
 				buffer.append("crop=true ");
 				//TODO: pass if possible the region to crop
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/env/ui/TaskBarManager.java
```diff
@@ -31,11 +31,7 @@
 import java.io.File;
 import java.io.InputStream;
 import java.io.InputStreamReader;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
```
dominikl: Was that intentionally?

dominikl:
Looks good. Just wondering about the modified import statements. Possible improvements: The default crop region is the whole image, in which case you'll get the error again (don't know from the top of my head if that could be somehow smartly adjusted too). Once the crop option is ticked because a large image was opened, then this becomes the default. Wonder if it should be unticked again automatically when a non-large image is attempted to be opened.

src/main/java/org/openmicroscopy/shoola/env/ui/TaskBarManager.java
jburel: No ide being annoying 😉

jburel:
I will look into setting the crop dimensions

jburel:
@dominikl crop options will be reset.
I have not found a way to specify the region in the string 
I tried ``X_Coordinate_1`` etc. but it is not picked 
cc @dgault

dominikl:
Looks good 👍 Tested with a local build again, works as expected. Automatically setting a reasonable crop region can be done in follow-up PR, if possible at all.

dominikl:
Still looks good 👍

Issue #94, opened 2019-11-14
Import wrong dataset


jburel:
push the "create" button in dataset creation dialog and push of "add to the queue" button are in normal but quick succession:

1. select a dataset, create new dataset A,click import;
2. while import 1 is running: select a dataset, create new dataset B, click import;
3. while import 1 is running: select a dataset, create new dataset C, click import;
After the import:
import 1 and import 2 are imported to dataset A; import 3 is imported to dataset C. Dataset B was created but empty. I had put a log messages to the LocationDialog::getImportSettings method and it seems so, that the LocationDialog doesn't get the right dataset destination if the time between creating a new dataset and add to queue event is very short and the connection to the server is slow

cc @sukunis @dominikl

jburel:
To avoid such situation and the current implementation of client and server
* either we create the object in the swing thread, this means that UI will block until the object is created and returned by the server
* or we do not create the object server-side when the user click ``New... `` The object will be created only when the import process starts similar to "New from folder"

The first option will be the easiest to implement but probably annoying for users when the server is slow

dominikl:
👍 for the first option. A bit a of delay wouldn't be unexpected at least.

joshmoore:
If there are other remote calls currently happening on the swing thread, then the creation of a dataset is unlikely to be an issue. If there _aren't_, then I'd be for trying to keep it that way since we certainly had in the past locked clients where the user had to force kill.

jburel:
there is no remote call on the swing thread, so hopefully the impact will be limited

jburel:
fixed in https://github.com/ome/omero-insight/pull/120

jburel:
https://github.com/ome/omero-insight/pull/121

PR #120, opened 2020-03-26
import wrong dataset

jburel:
This PR fixes https://github.com/ome/omero-insight/issues/94

The container is now created in the swing thread.
See the description in issue https://github.com/ome/omero-insight/issues/94 for testing instructions

cc @dominikl


jburel:
@dominikl this PR was not included in today's build so I removed it from standup

dominikl:
Looks good. Couldn't replicate the issue #94 with this PR. But I have to admit, although it's just a very short delay it feels a bit weird, like makes you question for a second if you actually clicked the button or not. Wonder if we could give some indication that something's happening. Wait cursor (but afaik we don't do that anywhere in Insight) or deactivate the "Create" button (in the `org.openmicroscopy.shoola.agents.util.ui.EditorDialog`)?

joshmoore:
Going back to https://github.com/ome/omero-insight/issues/94#issuecomment-604546275, J-m said:

> here is no remote call on the swing thread

If that's the case, then what I meant in https://github.com/ome/omero-insight/issues/94#issuecomment-604511096 was that I'd strongly encourage us not to introduce a remote call on the Swing thread. There be dragons if we can at all avoid it.

jburel:
I meant for that sequence. I think we might have one or 2 calls in the Swing thread but that will require confirmation. This is certainly not the norm

jburel:
@dominikl if we are in the swing thread, we won't be able to give feedback
As mentioned on the issue, option 2 is preferable. i will close this PR and see how we can sort that out.

PR #121, opened 2020-03-29
Import in wrong dataset

jburel:
Attempt to fix https://github.com/ome/omero-insight/issues/94

Check the scenarii described in the issue,

If a user creates a new dataset, the hierarchy will be reloaded at the end of the import similar to what we do for "new from folder".
This means that if the user tries to import in the dataset (to be) "created", another dataset with the same name will be created. Similar to what happen with "new from folder"
This is the less intrusive change I could see

I also notice a space missing in the tooltip of a control

cc @dominikl

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/FileSelectionTable.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/FileSelectionTable.java
@@ -86,7 +86,7 @@ class FileSelectionTable
 
 	/** Description of the <code>Add All</code> action.*/
 	private static final String TOOLTIP_BUTTON_ADD = "Add the selected files"+
-	        "to the queue.";
+	        " to the queue.";
 
 	/** Tooltip text for group column */
 	private static final String TOOLTIP_GROUP = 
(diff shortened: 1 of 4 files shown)


dominikl:
That's a good idea to use the same mechanism as "from folder"! Looks good, works fine 👍 Tried with a local build of the PR, because of travis error (just needs restarting I think).

jburel:
The dmg creation has been flaky lately

jburel:
Travis is green. So we can check again tomorrow in the daily build

jburel:
@pwalczysko are you happy with the strategy chosen to fix that issue?

pwalczysko:
> This means that if the user tries to import in the dataset (to be) "created", another dataset with the same name will be created. Similar to what happen with "new from folder"

@jburel Sorry, I am not getting this. The "it is just as "Create from Folder" simile is not helping me - I understand that sometimes the dataset will not get created quickly enough and the image ends up in another dataset (from previous import ?). Happens a lot during outreaches, yes, I was always thinking that the import did not work (unhelpfully, insight does not give you the id of the imported image, so I could never find the image ever again).

Could you please explain without using the "Create from folder"  motive in the explanation ?

pwalczysko:
Okay, understood after a chat.

I think this strategy is okay. Never understood why the Dataset is created immediately upfront, even as the import maybe does not succeed later, so this makes more sense.

jburel:
Thanks @pwalczysko

Issue #48, opened 2019-06-24
ImageJ plugin - channels


jburel:
When adding the "Split channels" option to the Crop workflow in ImageJ, see for example Fiji section, crop workflow of https://downloads.openmicroscopy.org/presentations/2019/Dundee-April/Dundee_April_2019_Day2.pdf

The channels names are not preserved correctly when the crop is imported back to OMERO.

   - lets have an image in OMERO with channels A, B, C
   - lets use the insight-ij plugin to open this image in Fiji and when in B-F dialog
   - lets select the "Split channels" checkbox
   - in Fiji, draw a rectangle on the image from channel C, then Image > Crop
   - Select Plugins > OMERO > Save image to OMERO and import the cropped image
   - in OMERO, observe that the crop is having one channel only (expected), but the name of this channels is "A" (not expected, "C" shoiuld be the name)

jburel:
see https://github.com/ome/omero-insight/pull/122

jburel:
The first part of the problem has been fixed in #122. 
We use the title of the ImagePlus as the name of the image
The second part of the issue is a problem with the BioFormats export plugin.
cc @dgault

PR #122, opened 2020-03-30
Fiji import image

jburel:
* Keep the title of the image plus.
This will keep the name of the channel in the final image name 
see https://github.com/ome/omero-insight/issues/48

cc @dominikl @pwalczysko

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/model/FileObject.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/model/FileObject.java
@@ -223,6 +223,7 @@ public class FileObject
             return ((File) file).getName();
         } else if (file instanceof ImagePlus) {
             ImagePlus img = (ImagePlus) file;
+            IJ.log(img.getTitle());
             return img.getTitle();
         }
         return null;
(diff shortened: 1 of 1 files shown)


dominikl:
Doesn't seem to work for me, the image still always has the name of the first channel.

jburel:
Could you indicate the image that you are using?

jburel:
Did you import the "selected window"?

dominikl:
Sorry, I think I used the wrong build earlier. Tried again with the correct, latest build but still have the same outcome. The channel name of the exported image is always 'Red' (used a normal RGB jpeg image as test), see merge-ci, user-3, dataset 'bee'.

jburel:
Clarification:
* The image name change works
* What is not working, the name of the channel is always the one of the first channel i.e. index 0

jburel:
The ome.tiff from the imageplus window is generated by B-F exporter plugin.
There is no obvious way to set the correct name. @dgault is checking if there is a possible workaround

jburel:
There is currently no option in B-F plugin to set the channel name. One option could be to "edit" the generated ome-tiff before import

jburel:
@dominikl @pwalczysko ok to integrate this PR, the "channel" name should be handled at B-F level.
Handling it here, will require a significant amount of work.

dominikl:
Sure, lgtm 👍

PR #123, opened 2020-03-30
bump packager version

jburel:
Test the new version of the plugin.
This means that the pom should no longer contain repository entries

To test:
 * check the pom and make sure it does not contain repository tag
 * travis is green

--exclude

cc @sbesson


dominikl:
I don't know what `org.openmicroscopy.additional-repositories` is, but the bump to version 5.5.2 works fine.

jburel:
This is part of the plugin to package app

PR #125, opened 2020-03-30
Remove ant build system

jburel:
We have not switched to gradle and we use travis to create the various executables
Those files are no longer necessary


.gitignore
```diff
@@ -9,6 +9,7 @@ out
 
 # Gradle cache
 .gradle
+gradle
```
joshmoore: What's creating this file?

jburel: If you want to test with another version of gradle than the one you have installed e.g. ``gradle wrapper --gradle-version=5.2.1``
I now personally use different conda env but this is not always the case

dominikl:
Works fine 👍

jburel:
nothing depends on ivy/ant anymore.

Issue #25, opened 2019-05-06
bin/omero-importer


jburel:
bin/omero-importer does not work
```Could not find or load main class org.openmicroscopy.shoola.Main```

**Comment by mtbc** [comment]
Still true for 5.5.8.

**Comment by PaulVanSchayck** [comment]
Still true for 5.5.9.

In case the cause was unclear. The startup scripts misses the contents of the CLASSPATH variable. After unzipping. Look in file `bin/omero-importer` file. The `CLASSPATH=` is empty.

When starting with empty `CLASSPATH`, you get the following error
```
$ ./bin/omero-importer 
Error: Could not find or load main class org.openmicroscopy.shoola.Main
```

Copying over the contents of the variable from the insight startup script fixes the issue.

jburel:
@PaulVanSchayck are you the importer? The reason I am asking is that we are considering not building the artifacts anymore

PaulVanSchayck:
Haha.. you've been considering that for years, haven't you? Every time I report an issue with the importer I get told exactly this :P.

We like the simple GUI. Especially because we tell users to use the webclient mostly, and hardly anyone uses the insight client. So we have only the importer installed on our microscopes and workstations.

jburel:
See https://forum.image.sc/t/omero-desktop-client-omero-insight/35234
The Desktop client will stay e.g. insight.dmg, we were just considering not bundling the importer as a standalone app.
If it is useful, we will keep bundling it.

PaulVanSchayck:
Hmm I see..

Would it be an option to include the `omero-importer` startup script in the insight artifact? Or as command line option to the `insight` startup script? Or do you think that's confusing?

jburel:
I have fixed the problem.
We did not hear that people still needed and it seems that you still do so we will keep building the artifact
I have also automated the build via travis so it should not be a problem.

jburel:
see https://github.com/ome/omero-insight/pull/127

jburel:
Fixed in 5.5.10

PR #127, opened 2020-04-10
Bin importer

jburel:
This fixes the issue described in https://github.com/ome/omero-insight/issues/25

To test locally
* run ``gradle build``
* unzip OMERO.importer.zip
* run ``bin/omero-importer``

or download the OMERO.importer from daily build and unzip

### DIFF
--- a/buildSrc/src/main/groovy/org/openmicroscopy/DistributePlugin.groovy
+++ b/buildSrc/src/main/groovy/org/openmicroscopy/DistributePlugin.groovy
@@ -29,6 +29,7 @@ import org.gradle.api.distribution.Distribution
 import org.gradle.api.distribution.DistributionContainer
 import org.gradle.api.distribution.plugins.DistributionPlugin
 import org.gradle.api.file.CopySpec
+import org.gradle.api.plugins.ApplicationPlugin
 import org.gradle.api.plugins.JavaPlugin
 import org.gradle.api.tasks.TaskProvider
 import org.gradle.api.tasks.application.CreateStartScripts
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/org/openmicroscopy/DistributePlugin.groovy
```diff
@@ -77,17 +78,21 @@ class DistributePlugin implements Plugin<Project> {
...
     }
 
     private TaskProvider<CreateStartScripts> addImporterCreateScriptsTask() {
         project.tasks.register(TASK_IMPORTER_START_SCRIPTS, CreateStartScripts, new Action<CreateStartScripts>() {
+            CreateStartScripts insight_css = project.tasks.named(ApplicationPlugin.TASK_START_SCRIPTS_NAME, CreateStartScripts).get()
+
             @Override
             void execute(CreateStartScripts css) {
-                css.mainClassName = InsightBasePlugin.MAIN_INSIGHT
+                css.mainClassName = InsightBasePlugin.MAIN_INSIGHT+" containerImporter.xml"
```
joshmoore, line 90: @jburel : I don't understand this change. Why is the XML file being added to `mainClassName`?

jburel: I can't find a way in ``CreateStartScripts`` to pass argument.
The mainClassName is then added to a bash command: ``JVMARG mainClassName etc.``
An other option could be to do bin/omero-importer containerImporterXml
but I opted to have the same approach that the one for insight

dominikl:
👍 Works fine, both for Insight and Importer.

buildSrc/src/main/groovy/org/openmicroscopy/DistributePlugin.groovy
jburel: another option could be to add another main class but this one, which is probably not the most elegant, is the less intrusive and require the less changes. Extending the gradle plugin will be the best option but it will require major work
This is very low priority

joshmoore: Understood. Let's go with it, keeping in mind that it's confusing and probably brittle.

jburel: I will add a TODO. The option might be added to the gradle plugin in the future

Issue #131, opened 2020-05-05
Move other users' data fails


dominikl:
Given one has the respective permissions, it should be possible to move other users' data, e.g. Images into another Dataset. This fails silently in Insight, but works as expected in Web and CLI. See forum post: https://forum.image.sc/t/query-for-datasetimagelink-id/37206/4

dominikl:
Problem fixed, but I leave the issue open for the RFE mentioned in the PR https://github.com/ome/omero-insight/pull/133#issuecomment-625108884 .

dominikl:
Error message fixed too: #141

PR #133, opened 2020-05-06
Remove canLink method

dominikl:
Removes the `canLink` method, which only checks if the dragged object belongs to the logged in user, and if it doesn't prevents the DnD action. I wonder why this method was there, because one doesn't necessarily have to be the owner of the object in order to be allowed to move it (see https://github.com/ome/omero-insight/issues/131 ).

**Test**: Check the scenario mentioned in the above issue now works.


jburel:
* User can move his/her own data. The checks are correct
* User with suitable permissions can move other users' data.

jburel:
One RFE. The message could be improved to "you do not have the correct permissions" in
This should be done in a follow-up PR.

Issue #132, opened 2020-05-06
Can not replace existing OMERO.script


emilroz:
This behaviour is different from the current Web implementation. OMERO.web replaces the existing script while Insight show the following error message:

```
java.lang.Exception: omero.gateway.exception.DSAccessException: Cannot access data. 
Cannot upload the script: Example_Script.py.
	at org.openmicroscopy.shoola.env.data.OMEROGateway.handleException(OMEROGateway.java:771)
	at org.openmicroscopy.shoola.env.data.OMEROGateway.uploadScript(OMEROGateway.java:6373)
	at org.openmicroscopy.shoola.env.data.OmeroImageServiceImpl.uploadScript(OmeroImageServiceImpl.java:1787)
	at org.openmicroscopy.shoola.env.data.views.calls.ScriptUploader$1.doCall(ScriptUploader.java:74)
	at org.openmicroscopy.shoola.env.data.views.BatchCall.doStep(BatchCall.java:144)
	at org.openmicroscopy.shoola.util.concur.tasks.CompositeTask.doStep(CompositeTask.java:226)
	at org.openmicroscopy.shoola.env.data.views.CompositeBatchCall.doStep(CompositeBatchCall.java:126)
	at org.openmicroscopy.shoola.util.concur.tasks.ExecCommand.exec(ExecCommand.java:165)
	at org.openmicroscopy.shoola.util.concur.tasks.ExecCommand.run(ExecCommand.java:276)
	at org.openmicroscopy.shoola.util.concur.tasks.AsyncProcessor$Runner.run(AsyncProcessor.java:91)
	at java.lang.Thread.run(Thread.java:748)
Caused by: omero.ApiUsageException
    serverStackTrace = ""
    serverExceptionClass = ""
    message = "Path already exists: \Desktop\Example_Script.py
               Use editScript to modify existing official scripts."
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at IceInternal.BasicStream.createUserException(BasicStream.java:2785)
	at IceInternal.BasicStream.access$300(BasicStream.java:14)
[...]

PR #134, opened 2020-05-06
Fix bug in script upload

dominikl:
Fixes a bug in the script upload functionality. The bug prevented scripts from being recognized as already existing on the server, so the uploadScript call failed. See https://github.com/ome/omero-insight/issues/132 .

**Test**: Check the workflow mentioned in the issue now works.


jburel:
With this PR, it is possible to re-upload an existing script.

PR #138, opened 2020-06-23
handle case with no dataset

jburel:
Handle the case when the dataset is not set.

See https://www.openmicroscopy.org/qa2/qa/feedback/29385/

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/model/ImportableFile.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/model/ImportableFile.java
@@ -153,10 +153,7 @@ public class ImportableFile
 	 */
 	public boolean isDatasetCreated()
 	{
-		if (dataset == null || dataset.getId() < 0) {
-			return false;
-		}
-		return  true;
+		return dataset != null && dataset.getId() >= 0;
 	}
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/env/data/model/ImportableFile.java
```diff
@@ -152,7 +153,10 @@ public FileObject getOriginalFile()
 	 */
 	public boolean isDatasetCreated()
 	{
-		return dataset.getId() >= 0;
+		if (dataset == null || dataset.getId() < 0) {
+			return false;
+		}
+		return  true;
```
dominikl, line 159: Could be simplified to `return dataset != null && dataset.getId() > -1`.

dominikl:
Thanks JM. The if-else isn't strictly necessary, but looks fine.

src/main/java/org/openmicroscopy/shoola/env/data/model/ImportableFile.java
jburel: true

dominikl:
👍 Good to merge.

PR #140, opened 2020-06-29
fix bug in MapAnnotationObject

sukunis:
fix bug in MapAnnotationObject


jburel:
@sukunis thanks for the quick fix. 
We are planning a release probably later this week so the fix will be included

jburel:
Merging so it can be included in the coming release

PR #141, opened 2020-06-30
Just change error message

dominikl:
Just a tiny PR which changes the error message, as suggest by @jburel in https://github.com/ome/omero-insight/pull/133#issuecomment-625108884 .


jburel:
Thanks

Issue #49, opened 2019-06-24
Bug: Insight import from root folder


jburel:
Insight appears to have trouble when importing files that are at the root of a mount if we try to create a new dataset from the local folder name. We have a server exception from a local Windows user at https://www.openmicroscopy.org/qa2/qa/feedback/27121/ (from their home on DIF) and locally on Linux it gives me an NPE.

mtbc:
I think this got fixed. Can try to dig up the PR. Do you remember @dominikl?

mtbc:
Oh sorry, I was remembering the card, I didn't realize this was the same but migrated! Ignore me. :smiley:

PR #145, opened 2020-07-01
Return root folder 'name' instead of empty string

dominikl:
Returns `/` (or `\\` on windows) as folder name, if an image from a filesystem root directory is imported.
Fixes https://github.com/ome/omero-insight/issues/49 . Without PR, if you choose "Dataset from folder" in the import dialog: Image goes to Orphaned Images, if no Project is selected (crash if a project is selected). With PR: A dataset with name `/` (`\\` respectively) is created for the image.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/model/FileObject.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/model/FileObject.java
@@ -31,6 +31,7 @@ import java.io.PrintWriter;
 import java.io.StringReader;
 import java.io.StringWriter;
 import java.net.URL;
+import java.nio.file.Path;
 import java.util.ArrayList;
 import java.util.List;
 
@@ -43,6 +44,7 @@ import loci.formats.codec.CompressionType;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.FilenameUtils;
 import org.openmicroscopy.shoola.util.CommonsLangUtils;
+import org.openmicroscopy.shoola.util.ui.UIUtilities;
 import org.w3c.dom.Document;
 import org.w3c.dom.NamedNodeMap;
 import org.w3c.dom.Node;
@@ -550,6 +552,8 @@ public class FileObject
                 parentFile = f.getParentFile();
                 if (parentFile == null)
                     return null;
+                if (parentFile.toPath().getNameCount() == 0)
+                    return UIUtilities.isWindowsOS() ? "\\\\" : "/";
                 return parentFile.getName();
             }
             return f.getName();


pwalczysko:
What will happen if someone has no right to access to that root folder ? We had such issues with ns users.

dominikl:
I don't think that's an issue for that particular case. If you don't have read access to the root directory, then you can't even select an image from there to import in first place.

jburel:
This certainly fixes the problem described in #49.
The image is no longer imported as "orphaned" and when a project is selected a dataset ``/`` is created.

The problem is now when importing from "root". A force quit is required when the import is successful.
Screenshot 2020-07-02 at 11 45 59

dominikl:
Thanks @jburel . Strange, I'll see where that comes from. There's just no such thing as a simple bug fix without side effects in Insight :-)

pwalczysko:
> If you don't have read access to the root directory, then you can't even select an image from there to import in first place.

Not exaclty. SLS has a system where the directories are visible only up to a certain level for rank-and-file users.

For example

``/Big-school-dir/sls-microscopy/users/user-1``

In this setup the users will have access only to each to their own directory (user-1 will have access only to user-1 directory etc.).  The directories users, sls-microscopy etc. are hidden for them. I think there is also a trick which hides the particulars of the directory ``user-1`` to the real user-1. Then, if user-1 has an image directly under ``user-1`` directory, and uses ``New from folder`` -> Insight crashes. This si a field observation, not a hypothesis.

jburel:
@pwalczysko does it crash with this PR?
I certainly got a crash in one scenario without this PR

dominikl:
@jburel Can't replicate the problem, works fine for me.

@pwalczysko I don't know how to replicate that. Still can't imagine how that can be a problem. Insight must be able to resolve the path of the image, otherwise it wouldn't couldn't present it in the import dialog in first place.

pwalczysko:
>  otherwise it couldn't present it in the import dialog in first place.

Well, I saw it otherwise. In the case I am talking about, insight was fine to present the file. But it was not fine when it tried to find out the name of the enclosing folder. Then it crashed. I also do not know how to replicate it. There must be a way to hide parts of the parh, SLS is doing it for their users. I think it could be replicated on ``diff`` or other SLS servers.

pwalczysko:
If https://github.com/ome/omero-insight/pull/145#issuecomment-653438562 is not possible to replicate (I am afraid I would not be able to replicate it myself, even if I would get an access to diff server, I am too powerful a user and can see the whole path), I am sure that this PR is improving things, so do not hold it for that, thanks.

dominikl:
Thanks Petr. It sounds like it might be fixed with this PR too. I'll have a look how I could replicate a scenario like that and give it a test.

dominikl:
I think this issue is fixed by this PR too. Mounted a "shared folder" in the virtualbox (`\\VBoxSvr`) in Windows10 VM (which afaik works via nfs under the hood too) and imported a file from the 'root' directory. Got imported into new dataset `\\` like expected.
Screen Shot 2020-07-03 at 10 30 05
(left insight, right windows explorer)

jburel:
@dominikl I did some tests again and noticed that
the "force quit" dialogs only pops up if a project is selected.

dominikl:
It looks like the issue is somewhere else. @pwalczysko found out that this happens against outreach server (always, no matter of mode of import!), but not against demo (which I've been using all the time).

dominikl:
--exclude (does this still work?) to check with tomorrows build if that's actually related to this PR or some other problem.

jburel:
It is not related to this PR
I selected an image in a folder (not root), select an existing project and ``New from folder``. 
and I had the "force quit" message again

pwalczysko:
> It is not related to this PR

Yes, this matches what I saw too.

jburel:
The problem described in #49 is fixed by this PR.
The "Force-quit" issue is not introduced in this PR. I will create a new GH.

jburel:
https://github.com/ome/omero-insight/issues/148

Issue #148, opened 2020-07-07
Bug: Import and project selection


jburel:
* User select an image to import
* Select a project
* "New from folder"
* Import the image
The import is successful, but the "Force Quit" dialog pops up when the user wants to quit

pwalczysko:
I actually can repeat also when no Project is selected. The difference for me is between outreach (bug is present) and demo (bug is not present) servers. Insight 5.5.11 on Mac

jburel:
interesting. i have checking against outreach.

jburel:
``countUploaded`` is not incremented.

dominikl:
Fixed in #149 .

PR #149, opened 2020-07-07
Import project

jburel:
This fixes the problem described in https://github.com/ome/omero-insight/issues/148

@dominikl, you introduced the ``hasParent`` check back in 2018 but I cannot remember the rationale behind it


dominikl:
Looks good. Tried different import scenarios, didn't notice any problems 👍

joshmoore:
Want to remove it completely then?

dominikl:
👍 Btw, I think it was introduced in 2013 https://github.com/ome/omero-insight/commit/5701269e15 , that was before my time ;-)

PR #163, opened 2020-09-11
handle null folder

jburel:
This PR fixes  a problem reported via qa system https://www.openmicroscopy.org/qa2/qa/feedback/29532/
cc @dominikl


src/main/java/org/openmicroscopy/shoola/env/ui/UserNotifierImpl.java
```diff
@@ -103,6 +103,12 @@
 	 * @return See above.
 	 */
 	private boolean canWriteInFolder(File folder) {
+		if (folder == null) {
+			notifyInfo("Download file",
+					"The selected folder cannot be null.\nPlease"
```
joshmoore, line 108: User probably doesn't know what a null folder means, eh?

jburel: They are clever :-)
I pushed a fix

mtbc (approved):
Seems a good enough guess given we've not managed to reproduce it!

PR #173, opened 2020-11-05
Actions to build on 3 os

jburel:
Build and run the tests on 3 os using GH actions
cc @sbesson

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -7,7 +7,7 @@ jobs:
   build:
     strategy:
       matrix:
-        java: [1.8]
+        java: [1.8, 11]
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,5 +1,5 @@
 ---
-name: Build with Gradle and run the tests
+name: Build
 
 on: [push]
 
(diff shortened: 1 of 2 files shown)


manics (approved):
Looks OK to me as a first step.

Does this build the installers too, or just the jars and launch script?

jburel:
Only the jars and the launch script.
For the dmg,`` javapackager`` that we rely on is not installed. It has also been removed from Java 11 (oracle).
The community is working on a new packager: ``jpackage``, so we might have to review our packaging strategy

joshmoore (approved):
Generally, :+1: I assume this doesn't change anything in our release process, right?

Also, the `name:` of the workflow can be used to have a badge URL if that's something you want to enable in later PRs.

jburel:
Travis is still used for the release, This is a first step.

.github/workflows/main.yml
```diff
@@ -0,0 +1,26 @@
+---
+name: Build with Gradle and run the tests
```
sbesson, line 2: Given its usage via the GitHub API e.g. badges, I have been leaning towards giving workflows concise single-word names and using the job name for additional verbosity - see e.g. https://github.com/ome/.github/blob/c8e7d8e8373f02cd956cd95e176f9aa2c0d41b74/workflow-templates/omero_plugin.yml#L10-L18

jburel: I will name it ``Build``

joshmoore:
:+1:

sbesson (approved):

PR #176, opened 2020-11-27
Gradle fix and publish

jburel:
Use the template file to use the correct gradle version

This PR fixes the GH actions
and add structure for releasing artifact 
See https://github.com/jburel/omero-insight/releases/tag/v5.5.15

To test the generation via GHA on your fork
 * change the version in build.gradle (dmg does not like snapshot) e.g. 5.5.16
 * push the tag to your fork
 * GHA will create the artifacts and create a release under your fork


dominikl:
👍 Seems to work fine: https://github.com/dominikl/omero-insight/releases/tag/v5.5.16 (I just forgot to commit the version change to 5.5.16 in the build.gradle).

jburel:
The msi did not get created
```
Failed to convert version component to int.
java.lang.NumberFormatException: For input string: "15-SNAPSHOT"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Integer.parseInt(Integer.java:580)
	at java.lang.Integer.parseInt(Integer.java:615)

```
b/c you kept snapshot. The error is not reported as a failure.
I might have to be smarter
Note this was the same problem with travis if I remember correctly

jburel:
@dominikl 
* could you repeat what you did i.e. keep snapshot and tag. The job should fail
* Remove snapshot and tag. The job should pass

dominikl:
Failed with snapshot: https://github.com/dominikl/omero-insight/actions/runs/410215701
Passed with proper version: https://github.com/dominikl/omero-insight/actions/runs/410245997
Looks good 👍

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,118 @@
...
+        run: gradle wrapper
+        if: ${{ env.gradle_version == '' }}
+      - name: Run commands
+        run: ./gradlew ${{ env.gradle_commands }}
+      - name: Prepare PATH
+        if: startsWith(matrix.os, 'windows')
+        shell: bash
+        run: echo "$WIX\\bin" >> $GITHUB_PATH
+      - name: Build Executable
+        if: startsWith(matrix.os, 'windows') && startsWith(github.ref, 'refs/tags') && matrix.java == '1.8'
```
sbesson, line 46: Is build time the primary reason for making this condition to tags? Or the SNAPSHOT vs non-SNAPSHOT?

.github/workflows/gradle.yml
```diff
@@ -0,0 +1,118 @@
...
+          fi
+          msi=(`find build/packaged/main/bundles -maxdepth 1 -name "*.msi"`)
+          if [ ${#msi[@]} == 0 ]; then
+              exit 1
+          fi
+        shell: bash
+      - name: Build Dmg
+        if: startsWith(matrix.os, 'macos') && startsWith(github.ref, 'refs/tags') && matrix.java == '1.8'
+        run: |
+          brew cask install zulu8
```
sbesson, line 69: Arguably all these steps could be combined into a standalone script e.g. `package_macos.sh` that gets executed by GitHub workflow but could be tested and modified in isolation.
Mostly a question, not a requirement for this PR

sbesson (commented):
All the changes generally make sense in order to produce packages, upload them as artifacts and combine them into release assets. A few questions inline but none is really a blocker to get this in.

Comparing https://github.com/dominikl/omero-insight/releases/tag/v5.5.18 generated with this PR with the last release https://github.com/ome/omero-insight/releases/tag/v5.5.14, is the removal of drop of `omero_ij-5.5.14-all.jar` on purpose?

.github/workflows/gradle.yml
jburel: this will increase the build time, that's why I have opted to make it on tag only

.github/workflows/gradle.yml
jburel: This could be refactored now that we have the build in place.
This could be done in a follow-up PR

jburel:
I must have dreamt that I thought the jar in the list of artifacts, it should be there. so another commit is required

jburel:
@sbesson This should now be fixed see https://github.com/jburel/omero-insight/releases/tag/v5.5.16

sbesson (approved):
Everything looks good with the last commit. I have not tested any of the artifacts so this might be something we will need to look at ahead of the next OMERO.insight release.

dominikl:
Sorry for the delay @jburel . Tested all artifacts of https://github.com/dominikl/omero-insight/releases/tag/v5.5.18 which were generated with the changes of this PR. All worked fine, except the Insight msi build. By default it wants to install itself in the Omero.importer directory. If you change the path during installation though it works fine too.

jburel:
Could you try with the msi using the release  since I reckon it is a problem with the gradle build itself and not the switch to actions?

dominikl:
Ok, this is really strange. I tried again and now like one would expect installation location is `C:\Program Files\OMERO.importer\` and `C:\Program Files\OMERO.insight\`. Before it was some weird `...AppData\Local\OMERO.importer\` location for both Insight and Importer.

jburel:
Thanks @dominikl

PR #183, opened 2021-01-12
MDE: Feature multiselection cb

sukunis:
add multiselection combobox as available inputfield type (define type="CheckComboBox")


src/config/mdeConfiguration.xml
```diff
@@ -30,6 +30,8 @@
                          Unit="mm" Value="millimeter value" Visible="true" />
                 <TagData DefaultValues="Value1,Value2,Value3" Name="Tag of Type ComboBox" Type="ComboBox"
                          Unit="" Value="Value1" Visible="true" />
+                <TagData DefaultValues="Value1,Value2,Value3" Name="Tag of Type CheckComboBox" Type="CheckComboBox"
```
dominikl, line 33: Just wondering if the Type should be changed to `CheckedComboBox` too?

src/config/mdeConfiguration.xml
```diff
@@ -380,6 +382,8 @@
                          Unit="nm" Value="" Visible="true" />
                 <TagData DefaultValues="Value1,Value2,Value3" Name="Tag of Type ComboBox" Type="ComboBox"
                          Unit="" Value="Value1" Visible="true" />
+                <TagData DefaultValues="Value1,Value2,Value3" Name="Tag of Type CheckComboBox" Type="CheckComboBox"
```
dominikl, line 385: Same here.

src/config/mdeConfiguration.xml
sukunis: To be honest: I don't quite understand what you mean.

dominikl: I was wondering if this should match the class name, which is `CheckedComboBox.java`.

dominikl:
👍 Looks good to me.

PR #185, opened 2021-01-13
MDE: fix unknown inputField type

sukunis:
create textfield if given type for inputField is unknown


dominikl:
Looks good to me 👍

PR #191, opened 2021-01-20
MDE: Feat filter required

sukunis:
If the "Show only required" check box is selected, only the content of objects that contain required tags will be displayed.


sukunis:
Sorry, there is a bug in this feature. I will fix it!

dominikl:
Looks good to me 👍

Issue #189, opened 2021-01-14
IllegalComponentStateException in ScriptingDialog


dominikl:
See [QA 29981](https://www.openmicroscopy.org/qa2/qa/feedback/29981/)
```
java.awt.IllegalComponentStateException: component must be showing on the screen to determine its location
	at java.desktop/java.awt.Component.getLocationOnScreen_NoTreeLock(Component.java:2101)
	at java.desktop/java.awt.Component.getLocationOnScreen(Component.java:2075)
	at java.desktop/javax.swing.JPopupMenu.show(JPopupMenu.java:940)
	at org.openmicroscopy.shoola.agents.util.ui.ScriptingDialog$1.mouseReleased(ScriptingDialog.java:387)
	
```

PR #194, opened 2021-01-26
Prevent potential IllegalComponentStateException

dominikl:
Only pop up the dialog when the component is visable. Fixes https://github.com/ome/omero-insight/issues/189 . 

**Test**:
Code review only, don't know how to reproduce the issue, must be very intermittent.


jburel:
Change makes sense

Issue #181, opened 2021-01-08
ImporterUI: IndexOutOfBoundsException


dominikl:
See [29794](https://www.openmicroscopy.org/qa2/qa/feedback/29794/)
```
java.lang.IndexOutOfBoundsException: Index: 0, Tab count: 0
	at javax.swing.JTabbedPane.checkIndex(JTabbedPane.java:1768)
	at javax.swing.JTabbedPane.setSelectedIndex(JTabbedPane.java:589)
	at org.openmicroscopy.shoola.agents.fsimporter.view.ImporterUI.selectChooser(ImporterUI.java:509)
	
```

Orohena:
Hi !
I am the one who raised that issue.
I can now explain how it happen if that help solving the issue:
In Omero insight, I clicked on the "Open the Importer" button.
Because I couldn't find what I was looking for, for some reason, I did a right click on the "tab" that say "Select the data to Import" which opened 3 options "Close", "Close Others", "Close All". I clicked "Close" and Omero crashed.

Hope it will help you understand the issue, anyway, I don't see why someone would try to Close that tab. Even I still don't understand why I did that :)

Flavie

dominikl:
Thanks @Orohena , indeed that's very helpful for locating and fixing the bug!

dominikl:
[30013](https://www.openmicroscopy.org/qa2/qa/feedback/30013/)

PR #195, opened 2021-01-26
Importdialog closeable

dominikl:
Prevent to close the ImportDialog tab. Fixes https://github.com/ome/omero-insight/issues/181 .

**Test**:
Right click on the "Select data to import" tab. Without the fix the "Close, Close All, etc." dialog will show up. If you close the tab via this dialog, close the importer dialog and open it again, it'll crash with the IllegalComponentStateException. With the fix you shouldn't have the possibility to close this tab.
Screen Shot 2021-01-26 at 15 28 02

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2018 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2021 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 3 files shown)


jburel:
* The right-click menu is no longer available for the ``Select data to import`` tab
* If I import an image, a new tab is added: The menu is then available as expected.
* If you select ``Close Others``, the ``Select data to import`` tab is removed and cannot be brought back
* If you select ``Close all``, the ``Select data to import`` tab is removed leading to an exception when re-opening the importer.
* the  ``Select data to import`` should never be deleted. The implementation of the close method also needs to be reviewed.

jburel:
* If you select ``Close Others``, the ``Select data`` to import tab is not removed
* If you select ``Close all``, the ``Select data to import tab`` is not removed but the window does not have the focus. This will be confusing for the user. The user needs to click on the tab to bring it back
Screenshot 2021-02-15 at 10 13 41

jburel:
The problem reported above https://github.com/ome/omero-insight/pull/195#issuecomment-779126725 is now fixed.
Merging

Issue #178, opened 2020-12-21
Search: Date to String cast error


dominikl:
Have to check if that's a swingx internal error or if it can be prevented in Insight.

[QA 29759](https://www.openmicroscopy.org/qa2/qa/feedback/29759/)
```
java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.String
	at java.text.Collator.compare(Collator.java:304)
	at javax.swing.DefaultRowSorter.compare(DefaultRowSorter.java:968)
...
```

From @sukunis 
The error can be reproduced. Here's what I did:
[1] search in insight for *.tif files
[2] search results are files with and without Acq date - let sort by Acq date

PR #196, opened 2021-01-26
Convert Date to String

dominikl:
Convert the Date column to String in the search result table. Fixes https://github.com/ome/omero-insight/issues/178 .

**Test**:
Instructions for replicating the bug see issue.


jburel:
Without that PR, the following error is thrown
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.ClassCastException: java.util.Date cannot be cast to java.lang.String
	at java.text.Collator.compare(Collator.java:304)
	at javax.swing.DefaultRowSorter.compare(DefaultRowSorter.java:968)
```

Issue #190, opened 2021-01-15
Allow to pass Ice config


jburel:
Allow support for ice config.
This will require change to the gateway to pass handle more info to "login credentials"

cc @dominikl

dominikl:
Does it need a change? If you use `LoginCredentials(String[] args)` constructor you can pass it through. The problem is that the `omero.client` itself has only these two options: Either you log in with username, password, etc. (then you can't  pass through any extra options) or you use the `client(String[] args)` constructor.

jburel:
I should have written "might require"
since I have not look at it in details

dominikl:
By always constructing the `String [] args` in the `LoginCredentials` it would be possible to add a convenience method to the gateway to make it easier to pass through addtional parameters. I'll add it to my draft PR https://github.com/ome/omero-insight/pull/182 .

joshmoore:
For what it's worth, `Ice.Communicator` (used in `omero.client`) should also read the contents of any `$ICE_CONFIG` file automatically, but at the moment, that looks to be disallowed by insight.

jburel:
insight reads the container.xml as the first argument if specified
so the logic needs to be adjusted

joshmoore:
I'm not sure what you're saying. That should be independent of an environment variable, no?

jburel:
the idea of the user was to do ``insight.exe ice.config ``, that will not work since insight will expect in that case ``insight.exe yourowncontainer.xml ``

joshmoore:
Right, but as a workaround perhaps `ICE_CONFIG=ice.config insight.exe` ? I was just hoping that would be easier.

dominikl:
Just adding a reference to the respective forum post: [Set Proxy in BlitzGateway](https://forum.image.sc/t/set-proxy-in-blitzgateway/43780/10)

jburel:
https://github.com/ome/omero-insight/pull/197 partially fixes the issue.
``ICE_CONFIG=pathToIceConfig`` is not supported yet

PR #197, opened 2021-01-26
Pass through commandline arguments

dominikl:
Allows to pass through command line arguments to `omero.client`, e.g. `--Ice.HTTPProxyHost=PROXY --Ice.HTTPProxyPort=PORT`. Fixes https://github.com/ome/omero-insight/issues/190 . I'm still not sure how to pass through an `ICE_CONFIG` environment variable. But afaik this shouldn't be necessary as you can just use the command line arguments with this PR.

**Test**:
To check if the arguments are taken into account just use something like `--Ice.HTTPProxyHost=1.2.3.4` and you shouldn't be able to connect. Remove it again and you can connect again.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/Main.java
+++ b/src/main/java/org/openmicroscopy/shoola/Main.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.Main
  *
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2010 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2021 University of Dundee. All rights reserved.
  *
  *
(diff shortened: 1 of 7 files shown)


jburel:
this PR handles the command line argument, it does not support the ``ICE_CONFIG=pathToIceConfig``.
This could be done in a follow-up PR

dominikl:
I should have phrased it differently. I actually have no explanation why the `ICE_CONFIG` is not taken into account. Insight doesn't seem to overwrite it anywhere, and it should be picked up but the `omero.client` automatically; no idea why that doesn't happen.

jburel:
This is an improvement. This fixes partially #190

Issue #192, opened 2021-01-26
NPE ThumbnailLoader


dominikl:
See [29989](https://www.openmicroscopy.org/qa2/qa/feedback/29989/), "searching for an image"

```
java.lang.Exception: java.lang.NullPointerException
	at org.openmicroscopy.shoola.env.data.views.calls.ThumbnailLoader.getThumbnailStore(ThumbnailLoader.java:302)
```

PR #198, opened 2021-01-27
Also catch NPEs in ThumbnailLoader

dominikl:
Catch all exceptions and use the error icon as thumbnails. Should fix https://github.com/ome/omero-insight/issues/192 (although I don't know why the Image's Pixeldata was `null` in that case).


jburel:
The change handles all the exceptions.
Unfortunately the source of the error is not recorded

dominikl:
True... I'll have a look, maybe I can at least log the error without throwing an exception (in follow-up PR).

PR #201, opened 2021-02-03
use gradle 6

jburel:
Build with gradle 6


dominikl:
Looks good to me. Also tested locally with gradle 6.8.2, work fine 👍

Issue #180, opened 2021-01-05
NPE LocationDialog.switchToSelectedGroup


dominikl:
Change group on import dialog, see [29780](https://www.openmicroscopy.org/qa2/qa/feedback/29780/)
```
java.lang.NullPointerException
at org.openmicroscopy.shoola.agents.fsimporter.chooser.LocationDialog.switchToSelectedGroup(LocationDialog.java:980)
	at org.openmicroscopy.shoola.agents.fsimporter.chooser.LocationDialog.itemStateChanged(LocationDialog.java:1786)
	at javax.swing.JComboBox.fireItemStateChanged(JComboBox.java:1223)
...
```

PR #202, opened 2021-02-15
Avoid potential NPEs

dominikl:
Avoid potential NPEs like https://github.com/ome/omero-insight/issues/180 in LocationDialog and ImportDialog.

**Test**:
I don't know how to replicate the NPE, but maybe just check that the Import workflow still works as expected.


jburel:
This could be due to a slow connection. 
The changes make sense

jburel:
tested the import workflow and did not notice any issue.
Merging

PR #213, opened 2021-03-22
trigger the mouse event linux

sukunis:
trigger the mouse event for object tree also under linux


src/main/java/org/openmicroscopy/shoola/agents/fsimporter/mde/components/view/ModuleTree.java
```diff
@@ -115,7 +115,16 @@ private void buildTree(DefaultMutableTreeNode elem)
 		tree.setEditable(false);
 		tree.getSelectionModel().setSelectionMode(TreeSelectionModel.SINGLE_TREE_SELECTION);
 		tree.addMouseListener(new MouseAdapter() {
-			public void mouseReleased(MouseEvent e) {
+			// triggers opoup event under linux
```
jburel, line 118: typo ``popup``

Issue #146, opened 2020-07-02
InvalidPathExceptions (\\?\D:\...)


dominikl:
Occassionally reported InvalidPathExceptions with strange paths beginning like `\\?\D:\...`:
[29247](https://www.openmicroscopy.org/qa2/qa/feedback/29247/) , [29251](https://www.openmicroscopy.org/qa2/qa/feedback/29251/) , [29312](https://www.openmicroscopy.org/qa2/qa/feedback/29312/)

They all came from Osnabrueck. As QA access is not sorted out for @sukunis yet , I just add the full exception here (with potential sensitive bits removed).

<details>
  <summary>Stracktrace</summary>
<p>

```
java.lang.Exception: Abnormal termination due to an uncaught exception.


java.nio.file.InvalidPathException: Illegal character [?] in path at index 2: \\?\D:\Freigegeben PAINT\Collaborations\....\some.tif


	at sun.nio.fs.WindowsPathParser.nextSlash(WindowsPathParser.java:212)


	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:111)


	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)


	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)


	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)


	at java.io.File.toPath(File.java:2234)


	at org.apache.commons.io.FileUtils.isSymlink(FileUtils.java:3107)


	at org.apache.commons.io.FileUtils.sizeOfDirectory0(FileUtils.java:2618)


	at org.apache.commons.io.FileUtils.sizeOf0(FileUtils.java:2641)


	at org.apache.commons.io.FileUtils.sizeOfDirectory0(FileUtils.java:2619)


	at org.apache.commons.io.FileUtils.sizeOfDirectory(FileUtils.java:2599)


	at org.openmicroscopy.shoola.env.data.model.FileObject.getLength(FileObject.java:450)


	at org.openmicroscopy.shoola.agents.fsimporter.chooser.FileElement.getFileLength(FileElement.java:135)


	at org.openmicroscopy.shoola.agents.fsimporter.chooser.FileElement.getFileLengthAsString(FileElement.java:145)


	at org.openmicroscopy.shoola.agents.fsimporter.chooser.FileSelectionTable.addFiles(FileSelectionTable.java:614)


[...]

PR #216, opened 2021-03-25
path error

jburel:
Select file to upload/import on Windows 10 
see https://github.com/ome/omero-insight/issues/146

cc @pwalczysko

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/util/ui/filechooser/GenericFileChooser.java
+++ b/src/main/java/org/openmicroscopy/shoola/util/ui/filechooser/GenericFileChooser.java
@@ -24,6 +24,7 @@ package org.openmicroscopy.shoola.util.ui.filechooser;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
 import java.io.File;
+import java.net.URL;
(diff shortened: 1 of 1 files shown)


pwalczysko:
On Windows 10, I cannot repeat the original problem. 
From the QA report, the problem seems to be an insertion of a question mark into the path of the image.
But no idea how to create such situation. Tried to create a shortcut and import the shortcut. This failed with a different error.

```
loci.formats.UnknownFormatException: Unknown file format: C:\Users\User\Desktop\test-jm.tif - Shortcut.lnk
	at loci.formats.ImageReader.getReader(ImageReader.java:202)
	at loci.formats.ImageReader.setId(ImageReader.java:844)
...`
```

jburel:
The error in that case is expected

pwalczysko:
> The error in that case is expected
> Did a dialog pop up?

No, no dialog popped up. Note that I was using the release insight.

dominikl:
Didn't get the "Cannot use shortcut..." dialog either. Just import fails with "Unknown file format". But doubt too that that's really the problem of https://github.com/ome/omero-insight/issues/146 (very weird path with `?` characters in the path name, etc.).

Issue #188, opened 2021-01-13
Olympus .vsi download bug


dominikl:
See [Olympus .vsi download bug using OMERO.insight](https://forum.image.sc/t/olympus-vsi-download-bug-using-omero-insight/47413).

When Insight downloads a vsi image it puts the files into a directory, this should not happen. Example from the dataset mentioned in the forum post:
Download of vsi image:
```
Brain_DAPI_GFAP-488_IBA3-Cy3_MBP-647_coronal10x_coronal-slideA_03.vsi
_Brain_DAPI_GFAP-488_IBA3-Cy3_MBP-647_coronal10x_coronal-slideA_03_/stack1/frame_t_0.ets
```
ends up in:
```
frame_t_0.ets/Brain_DAPI_GFAP-488_IBA3-Cy3_MBP-647_coronal10x_coronal-slideA_03.vsi
frame_t_0.ets/_Brain_DAPI_GFAP-488_IBA3-Cy3_MBP-647_coronal10x_coronal-slideA_03_/stack1/frame_t_0.ets
```

This is not the case with omero downloader. I assume because downloader tries to replicate the managed repository structure locally as close as possible.
I wonder how OMERO.web handles that case?

dominikl:
After some debugging I think I found the reason: For a multifile image Insight creates a directory for the files to be put in, using the image name as directory name. In some cases it picks up the vsi file first (correct) and in some cases it picks up the `frame_t_0.ets` file first (wrong). This is happening here: https://github.com/ome/omero-insight/blob/dfd2971de565b55d3c704e678feae56033584c63/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java#L3162 But I can't fix this in Insight, because it's an issue of the `Fileset` class: `getFilesetEntry(0)` or the equivalent `getPrimaryFilesetEntry()` return in these cases the ets file instead of the vsi.

CellKai:
Hi @dominikl thank your for looking into this issue (I wrote the post on image.sc).
Concerning this, we (@ehrenfeu @lguerard) think a possible solution from the user point of view would indeed be to NOT create the directory to put the files in in the first place . Actually, we think this would be even easier for the user to handle in the end. Do you think this could be a way to go for this issue?

dominikl:
👍  I think that makes sense. Just openend a PR for that.

CellKai:
super, that's awesome :) thank you very much!

CellKai:
[...]

PR #217, opened 2021-04-07
Don't create a directory for downloaded multifile image

dominikl:
Fixes https://github.com/ome/omero-insight/issues/188 by simply not creating a directory for a multifile image in first place, following the suggestion from @CellKai . Makes sense imo, or was there a specific reason for downloading the files of an image into a separate directory @jburel ?

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -3055,10 +3055,11 @@ class OMEROGateway
 	 * retrieve data from OMERO service.
 	 */
 	Map<Boolean, Object> getArchivedFiles(
-			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths)
+			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths,
+			boolean createMFIDirectory)
 		throws DSAccessException, DSOutOfServiceException
 	{
-		return retrieveArchivedFiles(ctx, file, image, keepOriginalPaths);
+		return retrieveArchivedFiles(ctx, file, image, keepOriginalPaths, createMFIDirectory);
 	}
 	
 	/**
@@ -3069,13 +3070,15 @@ class OMEROGateway
 	 * @param image The image to retrieve.
 	 * @param keepOriginalPaths Pass <code>true</code> to preserve the original folder
 	 *               structure.
+	 * @param createMFIDirectory Flag indicating to put multifile images into own directories
 	 * @return See above.
 	 * @throws DSOutOfServiceException If the connection is broken, or not logged in
 	 * @throws DSAccessException If an error occurred while trying to
 	 * retrieve data from OMERO service.
 	 */
 	private Map<Boolean, Object> retrieveArchivedFiles(
-			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths)
+			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths,
+			boolean createMFIDirectory)
 		throws DSAccessException, DSOutOfServiceException
 	{
 		List<?> files = null;
@@ -3142,6 +3145,9 @@ class OMEROGateway
 		List<String> notDownloaded = new ArrayList<String>();
 		String folderPath = null;
 		folderPath = file.getAbsolutePath();
+
+		Map<Fileset, String> filesetPaths = new HashMap<Fileset, String>();
+
 		Iterator<Entry<OriginalFile, Fileset>> entries = values.entrySet().iterator();
 		
 		while (entries.hasNext()) {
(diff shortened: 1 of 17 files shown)


jburel:
I cannot remember the rational behind the previous implementation.
This should be considered as a breaking change since it will change the expectation of the users

jburel:
After testing with master, I think the idea was to keep the original container of the image for some type of images.
Below are 2 screenshots. I downloaded the ``lionheart MRI DICOM``
(master)
Screenshot 2021-04-08 at 15 24 37
(PR) 
Screenshot 2021-04-08 at 15 24 54

Which images did you test with?

dominikl:
I tested with the vsi image mentioned on image.sc (you can find on the demo server, image id 149046) https://forum.image.sc/t/olympus-vsi-download-bug-using-omero-insight/47413 . The problem is that sometimes it works (ie. the created directory name makes sense) and sometimes it doesn't, it depends which file is picked from the set and taken to derive the directory name from.

jburel:
Which means that the proposed changes will work for the example you highlighted but not for the one I used. So it will be a regression for some users. We will have to be smarter, maybe a checkbox in the UI to keep the parent folder or not

jburel:
Which means that the proposed changes will work for the example you highlighted but not for the one I used. So it will be a regression for some users. We will have to be smarter, maybe a checkbox in the UI to keep the parent folder or not

dominikl:
Checkbox added. Not very pretty. Any ideas for improvements? 

Screen Shot 2021-04-14 at 16 56 29
(Tooltip says: "Put multifile images into a directory" )

By the way, I couldn't believe through how many classes and methods I had to pass that flag through!

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -3154,30 +3158,39 @@ IObject findIObject(SecurityContext ctx, String klassName, long id, boolean allG
...
-                path = path +"/"+ (of.getPath().getValue().replace(repoPath, ""));
-                // path should now be in the form
-                // DOWNLOAD_FOLDER/IMAGE_NAME[(N)]/X/Y/Z
-                // where X, Y, Z are image specific subdirectories
-                // for the single image/data files
-                File origPath = new File(path);
-                if (!origPath.exists())
-                    origPath.mkdirs();
+            	if (createMFIDirectory) {
+					// this will store multi file images within a subdirectory with
```
dominikl, line 3162: Sorry, I'm not quite sure what's happening here with the indent. Looks fine in Intellij, and it should use spaces instead of tabs.

pwalczysko:
The PR is suggesting not to create a directory for a vsi file by means of an optional checkbos.
Did some investigation of the status of things.

1. OMERO.web is downloading the vsi file from the image.sc post in a correct folder structure, see screenshot below, This downloaded vsi file and the whole structure can be reimported into OMERO.server (merge-ci, user-3, Project ID 36607). It consists of 2 images, one of them being the "macro image". This is as expected, as it is reflecting the original data layout on demo server.

OMERO web

2. OMERO.insight (release) is not downloading the same vsi file correctly, instead, it creates a subfolder with a wrong name, see screenshot below.
OMERO isnight-release

3. OMERO.insight from this PR with the "Create Dir" checkbox checked (default) is not downloading the same vsi file correctly either, making the same mistake as OMERO.insight release, see screenshot.
OMERO insight-PR-with-box-checked

4. OMERO.isnight from this PR with the "Create Dir" box unchecked is downloading the two files (vis and ets) and creates no directory. Nevertheless, this structure cannot be reimported into OMERO again, resulting in a loss of data (the macro image is lost, as it is not recognized by the vis reader I suppose, as the folder is missing). It is not possible to import this image separately either.
OMERO isnight-PR-with-box-unchecked
Screenshot 2021-05-03 at 11 45 50
```
loci.formats.FormatException: Could not find .vsi file.
	at loci.formats.in.CellSensReader.initFile(CellSensReader.java:634)
	at loci.formats.FormatReader.setId(FormatReader.java:1421)
	at loci.formats.ImageReader.setId(ImageReader.java:849)
	at
```

In summary, I see two problems here

1. The suggested fix to the problem reported on image.sc is leading to loss of data, even on this very image, whose download should be fixed by this PR in the first place. This strongly suggests that the non-dreation of the directory might and very probably is leading to other losses of data in other cases.
2. It is not clear from the button itself that the situation Ad. 1. above is going to occurr. Indeed, it is not very clear what that button is doing (unless you know this PR and the problematic upfront). I would guess at least a warining when unchecking that box that this might lead to loss of data would be appropriate.

I would suggest to come back to the image.sc issue and ask the users there if they considered the data loss situation in their suggestion as described in Ad 1. above. 
To move forward, I woudl suggest once more to study how OMERO.web is managing to download this file in a correct naming of the folder and try to implement the same logic in insight.

ehrenfeu:
Dear @pwalczysko,

thanks a lot for the detailed update and debugging attemtps. I guess what my colleague @cellkai meant in his image.sc forum post was **not** to skip the creation of folders entirely, but rather what he described as "packaging folders" (meaning to skip the **additional** creation of a "surrounding" folder for the multifile-download if that description explains it any better...).

The correct layout for a "VSI package" (.vsi plus companion files and folder structure) should look e.g. like this:

```
VSI_Olympus/
├── _Live-Trial1-SpinD_01_
│   └── stack1
│       └── frame_t_0.ets
└── Live-Trial1-SpinD_01.vsi
```

Does this help in any way?

Cheers,
Niko

pwalczysko:
> Dear @pwalczysko,
> 
> thanks a lot for the detailed update and debugging attemtps. I guess what my colleague @CellKai meant in his image.sc forum post was **not** to skip the creation of folders entirely, but rather what he described as "packaging folders" (meaning to skip the **additional** creation of a "surrounding" folder for the multifile-download if that description explains it any better...).
> 
> The correct layout for a "VSI package" (.vsi plus companion files and folder structure) should look e.g. like this:
> 
> ```
> VSI_Olympus/
> ├── _Live-Trial1-SpinD_01_
> │   └── stack1
> │       └── frame_t_0.ets
> └── Live-Trial1-SpinD_01.vsi
> ```
> 
> Does this help in any way?
> 
> Cheers,
> Niko

Dear Niko,
thank you very much for this, indeed, it does help a lot. Yes, you have it right imho, insight should not be skipping the folders which are crucial for the proprietary file formats like vsi to function properly. Unfortunately, at present, that is where we are at this moment in this PR. We will have to recosider the tactics  here with @dominikl and @jburel and come up with some solution, thanks again.
Petr

dominikl:
Doesn't solve the issue.

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -3154,30 +3158,39 @@ IObject findIObject(SecurityContext ctx, String klassName, long id, boolean allG
...
+					// the same name as the main image file
+					path = filesetPaths.get(set);
+					if (path == null) {
+						path = folderPath.endsWith("/") ? folderPath : folderPath + "/";
+						String imgFilename = set.getFilesetEntry(0).getOriginalFile()
```
joshmoore, line 3167: From our discussion today, does something else call `getOriginalFile().getPath()`?

ehrenfeu:
Hey guys, I'm sorry to bring up this topic again but it is a real bummer for us - it actually makes it very difficult for our VSI-users (and we're having two heavily used Olympus spinning disk systems...) to trust into OMERO as they don't get their data back in the expected format...

Is there anything we can do on this?

Sorry for being nasty,
~Niko

CellKai:
Hi @dominikl , thank you for picking up the issue again! I work in the same group as Niko.
OMERO.web does indeed handle the these cases correctly. However, using OMERO.web in the meantime as a workaround does not work, as there are other issues in OMERO.web that prevent the download of large image (>4 GB) files, see
- https://github.com/ome/omero-web/issues/241
- https://trello.com/c/HXHVwDXg/140-omeroweb-download-issue
We need to rely therefore on OMERO.insight to get our VSI-images out of OMERO.

dominikl:
Yes, that's true, forgot about the size download size limitation. At least I can take OMERO.web as reference.

Issue #221, opened 2021-05-05
Sanitiy checks in Import dialog


dominikl:
At the moment you can enter options which cause the import to fail, e. g. invalid pixel sizes: [QA 30140](https://www.openmicroscopy.org/qa2/qa/feedback/30140/) Should have some kind of basic sanity checks there.

pwalczysko:
I have a workflow which ends in 
``java.lang.IllegalArgumentException: values of Pixels.physicalSizeZ must be strictly positive"``

The steps to reproduce:

1. Fill in the X and Y with reasonable values (e.g. 0.1) and do not fill in Z at all 
2. Observe that import succeeds and the zutocomplete for Z is "1"
3. Fill in the X and Y with reasonable values, but also start filling in the Z. 
4. Delete the value you filled in into Z box completely.
5. The reasonable expectation now would be that the Z would be autocompleted with "1" as in Ad 2. above.
6. Instead, the autocomplete is "0.0" for the Z box, which is then failing on import with the error above.
7. I think it is fair to say that it is not about insight not checking, but rather actively suggesting a value which will later fail.

sbesson:
Looking at this workflow, I'd also like to flag that filling out a pixel size in the way described above can lead to many issues down the lines. Incorrect metadata is usually harder to deal with than missing metadata.

I would be in favor of completing disabling the value autocompletion in this particular case.

pwalczysko:
> Looking at this workflow, I'd also like to flag that filling out a pixel size in the way described above can lead to many issues down the lines. Incorrect metadata is usually harder to deal with than missing metadata.
> 
> I would be in favor of completing disabling the value autocompletion in this particular case.

Agree with the reasoning here, but we might be in danger of hampering usability like this.
Scenario (run through this myself as well as feedback from other users indicative to this)
User: "My images are not z-stacks, thus I do not consider z to be important. Even if it should be important, I have no idea what to fill in.  I ignore the z value and go ahead with import. But then, after a while, the import crashes (because OMERO insists on z value being filled in). I have to go back to the dialog and ponder what to fill in z (which I do not know...)."


This is a very real scenario  as documented in Tips & Tricks gif (the one I am talking about is not published yet, will pass it via email.), cc @dominikl @sbesson

pwalczysko:
[...]

PR #222, opened 2021-05-19
Prevent <=0 pixel sizes

dominikl:
Fixes https://github.com/ome/omero-insight/issues/221 by simply preventing values <=0 to be entered. The bigger issue about allowing `null` needs to be tackled on the API level, not within Insight.
Test: Just try to enter an invalid pixel value.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
@@ -56,6 +56,7 @@ import javax.swing.Box;
 import javax.swing.BoxLayout;
 import javax.swing.ButtonGroup;
 import javax.swing.Icon;
+import javax.swing.InputVerifier;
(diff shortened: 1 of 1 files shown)


pwalczysko:
the behaviour with this PR is

- if the value in the Z box is <=0 then the system autofills "1" (not into the box, but later in the import process, as verified in the UI after import)
- if the value in the Z box is empty, hen the system autofills "1" (not into the box, but later in the import process, as verified in the UI after import)
- if the user types some value in Z box and deletes it immediately, hen the system autofills "1" (not into the box, but later in the import process, as verified in the UI after import)
- if the value in the Z box is positive, then teh system respects this positive value as verified on the image after import

This is not solving the problem of the "magic 1" being passed into the Z pixelsize value, but it is most certainly an improvement of the user experience against the present state.

Ready to merge fmpov

dominikl:
Yes, that's the outstanding issue, that Insight can't pass `null` as z-value to the respective method which sets the pixelsizes, it needs to be a number > 0, hence the 1.

Issue #224, opened 2021-05-28
Cannot log in with leading or trailing whitespace, #, = or \ in your password


chris-allan:
When #197 was merged this resulted in the standard login parameters of username, password, host and port going through the `omero.client(String[])` constructor rather than the `omero.client(String, int)` constructor followed by invocation of `omero.client#createSession(String, String)`. The use of the `omero.client(String[])` constructor results in a conversion of these standard login parameters into Ice configuration properties. These configuration properties are subject to the special character semantics defined here:

 * https://doc.zeroc.com/ice/3.6/properties-and-configuration/configuration-file-syntax

Essentially, `#`, `\` and `=` have a special meaning and leading or trailing whitespace needs to be escaped. I think the simplest approach to resolve this would be to escape parameters when the arguments list is built in `UserCredentials#getArguments()` but it might be worthwhile to re-evaluate the current strategy altogether.

As a temporary workaround, users can escape leading or trailing whitespace, `#`, `\` and `=` themselves when entering the password into the OMERO.insight login dialog.

__Edit:__ Updated to reflect all the characters that need to be escaped.

/cc @emilroz, @erindiel

chris-allan:
It's probably not as common a problem as `#` and `=` but I believe leading or trailing whitespace also would need to be escaped if we are going to continue to use this constructor. Though I haven't personally tested it.

dominikl:
Maybe we simply have to revert  #197 instead of making it even more fiddly by starting to escape certain characters under the hood.

chris-allan:
With all the things that now need to be escaped, @dominikl, that might be the simplest course of action.

PR #227, opened 2021-06-02
Escape ice special characters for login

dominikl:
This should properly fix https://github.com/ome/omero-insight/issues/224 .

Test: Use 'pwtest' user with password ' test' (leading space!) and check the logs that the first login attempt works.


pwalczysko:
Works aas described. Was also able to reproduce the failure with the release insight.
Ready to merge fmpov

jburel:
Thanks 
Merging and I will prepare the release

PR #230, opened 2021-06-14
enable use of OMERO.importer with read-only server

sukunis:
enables use of OMERO.importer with read-only server for metadata annotations with OMERO.mde.
Could be useful to assist the user with IDR submissions or to export to other metadata and data formats using OMERO.mde.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
@@ -78,6 +78,9 @@ import javax.swing.filechooser.FileFilter;
 
 import loci.formats.gui.ComboFileFilter;
 
+import omero.gateway.Gateway;
+import omero.gateway.SecurityContext;
+import omero.gateway.exception.DSOutOfServiceException;
 import org.apache.commons.collections4.CollectionUtils;
 import org.jdesktop.swingx.JXTaskPane;
 import org.openmicroscopy.shoola.agents.fsimporter.IconManager;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openmicroscopy/shoola/agents/fsimporter/chooser/ImportDialog.java
```diff
@@ -1843,11 +1846,28 @@ public JButton getCancelImportButton() {
...
+
 		showMDEButton.setEnabled(enable);
 	}
 
+	/**
+	 * Helper function to use OMERO.importer with an server address of a read-only server (e.g. idr).
+	 * Check if users group is a non-read-only group.
+	 * @return
+	 */
+	private boolean canCreate() {
```
jburel, line 1861: I believe something like 
```
    public static boolean canCreate() {
        Boolean b = (Boolean) registry.lookup(LookupNames.CAN_CREATE);
        return b.booleanValue();
    }
```
at the agent level should do what you want 
see TreeViewerAgent for example

src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
```diff
@@ -564,7 +564,7 @@ public void connect(UserCredentials uc)
         if (!canCreate && LookupNames.MASTER_IMPORTER.equals(name)) {
             registry.getUserNotifier().notifyError("Read-only server",
                     "Cannot import on a read-only server!");
-            exitApplication(true, true);
+            //exitApplication(true, true);
```
jburel, line 567: We will need to carefully investigate the implication of this change

jburel:
The offers the ability to access mde but import is not enabled.

PR #232, opened 2021-06-16
bump gateway version

jburel:
Use the latest gateway version


jburel:
Tested indirectly via https://github.com/ome/omero-insight/pull/227

PR #235, opened 2021-07-06
Load thumbnails when read-only

jburel:
Check that you can view the thumbnail when connecting to IDR
The functionality should remain the same for other server

Problem noticed while testing a workflow for the upcoming course
cc @pwalczysko


dominikl:
Do you login as "public" user? I can't see any projects or screens when I log into IDR with Insight (as public user).

jburel:
public user then you need to use the group display

dominikl:
Looks good to me 👍 Could also replicate the problem with current Insight, no problem with merge build. Didn't realise earlier that you have to untick the "Display users" checkbox in order to see something.

pwalczysko:
Tested both ImageJ plugin and insight. Also, tested insight connecting to a training server. All works as expected.
Ready to merge fmpov

pwalczysko (approved):

PR #238, opened 2021-07-09
Fix NPE when components is null

jburel:
Handle case when there is no components
see https://www.openmicroscopy.org/qa2/qa/feedback/30324/


dominikl:
👍 Looks good to me.

Issue #210, opened 2021-03-04
javax.swing.JComponent: ArrayIndexOutOfBoundsException


dominikl:
[QA 30043](https://www.openmicroscopy.org/qa2/qa/feedback/30043) (and others). 
```
java.lang.ArrayIndexOutOfBoundsException: No such child: 3
	at java.awt.Container.getComponent(Container.java:336)
	at javax.swing.JComponent.rectangleIsObscured(JComponent.java:4390)
	at javax.swing.JComponent.paint(JComponent.java:1054)
...
```
Gets reported occassionally. Haven't found a way yet to reproduce it. And unfortunately the stacktraces never reveal which Insight component triggers the error.

dominikl:
[QA 30049](https://www.openmicroscopy.org/qa2/qa/feedback/30049/)

dominikl:
[30053](https://www.openmicroscopy.org/qa2/qa/feedback/30053/)

dominikl:
[30067](https://www.openmicroscopy.org/qa2/qa/feedback/30067/) Gettings these a lot now. Still don't know where/how this is triggered.

dominikl:
Maybe the bug is actually older... https://trac.openmicroscopy.org/ome/ticket/7250

dominikl:
[30137](https://www.openmicroscopy.org/qa2/qa/feedback/30137)

dominikl:
[QA 30143](https://www.openmicroscopy.org/qa2/qa/feedback/30143/)

dominikl:
Another one, this time with Java 14: [30157](https://www.openmicroscopy.org/qa2/qa/feedback/30157/)
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.ArrayIndexOutOfBoundsException: No such child: 3
	at java.desktop/java.awt.Container.getComponent(Container.java:350)
	at java.desktop/javax.swing.JComponent.rectangleIsObscured(JComponent.java:4426)
	at java.desktop/javax.swing.JComponent.paint(JComponent.java:1072)
...
```

jburel:
see https://www.openmicroscopy.org/qa2/qa/feedback/30244/

dominikl:
[...]

PR #240, opened 2021-07-16
Move UI update code in Swing thread

dominikl:
Attempt to fix https://github.com/ome/omero-insight/issues/210 . 
The error from the issue is typical if UI components have been removed from outside the Swing thread. In this code block a lot of UI updates, additions and removal of components is happening and it might be called asynchronosly. I suspect this could be the cause of the issue. Putting it into the Swing thread hopefully will fix the problem.

**Test**: Just test if this PR causes any issues for importing images. Test some imports, imports with just a few images, and imports with 2000+ images.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2018 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2021 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
```diff
@@ -893,88 +893,92 @@ private boolean hasAssociatedFiles() {
...
-			} else formatResult();
-		} else if (image instanceof Boolean) {
-			busyLabel.setBusy(false);
-			busyLabel.setVisible(false);
-			cancelButton.setVisible(false);
-			if (status.isMarkedAsCancel() ||
-					status.isMarkedAsDuplicate()) {
-				resultIndex = ImportStatus.IGNORED;
-				this.image = null;
+		SwingUtilities.invokeLater(new Runnable() {
```
jburel, line 896: import is missing leading to the failure

dominikl: Hm, I'm not sure about that... Do you know how to replicate the issue? I think it happens randomly when something goes out of sync with the UI thread.

jburel:
```
insight/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java:896: error: cannot find symbol

		SwingUtilities.invokeLater(new Runnable()
```

pwalczysko:
tested a simple import but for the 2000 images folder need at least an overnight run

pwalczysko:
The 2000 images import went smoothly as well with the build from merge-ci

The imported dataset is https://merge-ci.openmicroscopy.org/web/webclient/?show=dataset-43169 user-3. All images look good.


Screenshot 2021-09-01 at 12 21 14

dominikl:
Thanks @pwalczysko . Maybe we could release this at some point, and then check if we still get errors https://github.com/ome/omero-insight/issues/210 .

dominikl:
Ah, I guess would be good to bump it to the latest gateway as well.

jburel:
ok. Let's prepare a gateway too and release insight

Issue #243, opened 2021-07-27
Link in Insight points to old forum


dominikl:
The link in Insight still points to the old openmicroscopy.org forum.
![Screen Shot 2021-07-27 at 09 50 00](https://user-images.githubusercontent.com/6575139/127125524-cddc8aff-9aca-4faf-971d-eb0eb578ea2f.png)

PR #244, opened 2021-07-29
Point to image.sc forum

dominikl:
See title. Fixes https://github.com/ome/omero-insight/issues/243 .

### DIFF
--- a/src/config/container.xml
+++ b/src/config/container.xml
@@ -326,7 +326,7 @@
     <entry name="AboutFile">about.xml</entry>
     <entry name="HelpOnLine">https://help.openmicroscopy.org/</entry>
     <entry name="HelpOnLineSearch">https://help.openmicroscopy.org/search.html</entry>
-    <entry name="Forum">https://www.openmicroscopy.org/community/</entry>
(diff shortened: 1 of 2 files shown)


sbesson (approved):
Straightforward URL update to point to the image.sc forum. Is it critical enough to make available via a patch release or rather release it with another set of changes?

dominikl:
As nobody has noticed it so far, I think it probably doesn't require an immediate release.

src/config/container.xml
```diff
@@ -326,7 +326,7 @@
     <entry name="AboutFile">about.xml</entry>
     <entry name="HelpOnLine">https://help.openmicroscopy.org/</entry>
     <entry name="HelpOnLineSearch">https://help.openmicroscopy.org/search.html</entry>
-    <entry name="Forum">https://www.openmicroscopy.org/community/</entry>
+    <entry name="Forum">https://forum.image.sc/</entry>
```
sbesson, line 329: See https://github.com/ome/omero-py/pull/297#discussion_r681473522, for unification we might want to point at https://forum.image.sc/tag/omero-insight

PR #247, opened 2021-08-04
Point to forum, tag Insight

dominikl:
Follow up to #244 . Point to `omero-insight` forum tag. 
/cc @sbesson .


joshmoore (approved):
Thanks, @dominikl !

Issue #246, opened 2021-08-02
IllegalArgumentException: HistogramPlot


dominikl:
Selecting channels to see histograms in the graph pane of the ROI tool.

See [QA 30360](https://www.openmicroscopy.org/qa2/qa/feedback/30360/)
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.IllegalArgumentException: Mismatch between argument length
	at org.openmicroscopy.shoola.util.ui.graphutils.HistogramPlot.(HistogramPlot.java:115)
	at org.openmicroscopy.shoola.agents.measurement.view.GraphPane.drawHistogram(GraphPane.java:370)
...
```

PR #249, opened 2021-08-04
Don't throw Exception if there's no data to show

dominikl:
Don't throw Exception if there's no data to show, simply show empy histogram panel.
Fixes https://github.com/ome/omero-insight/issues/246 .

Replicate the issue: Open image with an ROI. Open ROI tool. Go to "Graph Pane". Select ROI. This should show the histogram. Then de-activate the channels in the Image viewer. Crash with IllegalArgumentException .
With the PR this should not crash, but just show an empy histogram, if the channels are deactivated.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/util/ui/graphutils/HistogramPlot.java
+++ b/src/main/java/org/openmicroscopy/shoola/util/ui/graphutils/HistogramPlot.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.util.ui.graphutils.HistogramPlot 
  *
   *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2015 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2021 University of Dundee. All rights reserved.
  *
  *
  * 	This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


pwalczysko:
I am afraid I cannot reliably repeat the bug on 5.5.19 insight. It works in most cases. The workflow which elicited the bug was

 - open an image with ROIs
 - open ROI tool
 - click on Graph Pane
 - see the histogram getting populated
 - switch off all the channels
 - switch all the channels back on
 - observe the error (but only in some 25% of repeats of this workflow)

pwalczysko:
I think considering the fact that the testing would not be conclusive because of https://github.com/ome/omero-insight/pull/249#issuecomment-893605409, and that the error, when it appears, seems to be caused by switching the channels back on (does not appear when the channels are off), we should have one more look at the code.
The situation for me now is that I cannot repeat at all on 5.5.19 release insight.

dominikl:
Sorry, now I figured out how to repeat the issue (and why you got it only in 25% of cases). You have to quickly switch the channels off and on, because the error happens when the channels are switched off/on while the "Analyzing shape" process is running. With the PR there's no error.

pwalczysko:
> Sorry, now I figured out how to repeat the issue (and why you got it only in 25% of cases). You have to quickly switch the channels off and on, because the error happens when the channels are switched off/on while the "Analyzing shape" process is running. With the PR there's no error.

Great, thanks, indeed, now I am having 100% repeat rate. Even one channel quick switch off and on is triggering the bug. I will go on now with testing of your fix.

pwalczysko:
The insight from https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/ does not have the problem anymore.
When the channel is quickly switched off and on, then the drawing of the histogram is not happening (and so obviously the bug would have been triggered), but with the insight with this PR, no bug happens. When I switch next time just any channel eitcher on or off (i.e. not on and off in rapid order, just one action), then the drawing of the histogram resumes as expected.

imho, this is acceptable in general, and certainly the PR fixes the full crash.

Ready to merge fmpov

PR #264, opened 2021-12-10
Simply download the OriginalFiles

dominikl:
Fixes  https://github.com/ome/omero-insight/issues/188, successor of closed PR #217. 

I'm very tempted to just keep it simple @jburel , ie. with this PR the download option will simply download the OriginalFiles and their directory structure. It won't create a separate directory or rename any files. If a file already exists, it will be skipped. If we make it more complex again, we'll run into all sorts of issues with special cases again.

/cc @ehrenfeu @CellKai 

(Note: Haven't tested it on Windows yet)

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
@@ -1,6 +1,6 @@
 /*
  *------------------------------------------------------------------------------
- *  Copyright (C) 2006-2020 University of Dundee. All rights reserved.
+ *  Copyright (C) 2006-2021 University of Dundee. All rights reserved.
  *
  *
  *  This program is free software; you can redistribute it and/or modify
@@ -3078,30 +3078,27 @@ class OMEROGateway
 			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths)
 		throws DSAccessException, DSOutOfServiceException
 	{
-		return retrieveArchivedFiles(ctx, file, image, keepOriginalPaths);
+		return retrieveArchivedFiles(ctx, file, image);
 	}
-	
+
 	/**
 	 * Retrieves the archived files if any for the specified set of pixels.
 	 *
 	 * @param ctx The security context.
 	 * @param file The location where to save the files.
 	 * @param image The image to retrieve.
-	 * @param keepOriginalPaths Pass <code>true</code> to preserve the original folder
-	 *               structure.
 	 * @return See above.
-	 * @throws DSOutOfServiceException If the connection is broken, or not logged in
 	 * @throws DSAccessException If an error occurred while trying to
 	 * retrieve data from OMERO service.
 	 */
 	private Map<Boolean, Object> retrieveArchivedFiles(
-			SecurityContext ctx, File file, ImageData image, boolean keepOriginalPaths)
-		throws DSAccessException, DSOutOfServiceException
+			SecurityContext ctx, File file, ImageData image)
+			throws DSAccessException
 	{
 		List<?> files = null;
 		String query;
 		try {
-		    IQueryPrx service = gw.getQueryService(ctx);
+			IQueryPrx service = gw.getQueryService(ctx);
 			ParametersI param = new ParametersI();
 			long id;
 			if (image.isFSImage()) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -3131,134 +3128,79 @@ IObject findIObject(SecurityContext ctx, String klassName, long id, boolean allG
...
-				while (j.hasNext()) {
-					FilesetEntry fs = j.next();
-					values.put(fs.getOriginalFile(), set);
+			for (Object tmp : files) {
+				Fileset fs = (Fileset) tmp;
+				String repoPath = fs.getTemplatePrefix().getValue();
+				for (FilesetEntry fse: fs.copyUsedFiles()) {
+					OriginalFile of = fse.getOriginalFile();
+					String dir = of.getPath().getValue().replace(repoPath, "");
+					File outDir = new File(file.getAbsolutePath()+"/"+dir);
```
jburel, line 3143: File.separator should be used

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -3131,134 +3128,79 @@ IObject findIObject(SecurityContext ctx, String klassName, long id, boolean allG
...
-					results.remove(f);
+		try {
+			RawFileStorePrx store = gw.getRawFileService(ctx);
+			store.setFileId(of.getId().getValue());
+
+			FileOutputStream stream = new FileOutputStream(out);
+			long size = of.getSize().getValue();
+
+			long offset = 0;
+			try {
```
jburel, line 3188: Code can be simplified using ``try-with-resources`` and remove the ``stream.close()`` statement below
```
			try (FileOutputStream stream = new FileOutputStream(out) {
....
}
```

pwalczysko:
I have a problem with the insight zip on https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/986/.
When I download it and try to unzip it, I get

```
unzip "OMERO.insight-5.6.3-SNAPSHOT(1).zip"
Archive:  OMERO.insight-5.6.3-SNAPSHOT(1).zip
  End-of-central-directory signature not found.  Either this file is not
  a zipfile, or it constitutes one disk of a multi-part archive.  In the
  latter case the central directory and zipfile comment will be found on
  the last disk(s) of this archive.
unzip:  cannot find zipfile directory in one of OMERO.insight-5.6.3-SNAPSHOT(1).zip or
        OMERO.insight-5.6.3-SNAPSHOT(1).zip.zip, and cannot find OMERO.insight-5.6.3-SNAPSHOT(1).zip.ZIP, period.
```

If I try to download and unzip the imagej zip from the same build, or, if I try to download an insight zip from older build like https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/984/, I do not get this error. Thus, it looks like the most recent build is not building the insight zip correctly.
This also blocks me from further testing.

pwalczysko:
Deleted the comment, as I was able to download and unzip on subsequent attemps.

pwalczysko:
Test 1:
1. Download one Dicom image from https://outreach.openmicroscopy.org/webclient/?show=image-13402 
2. Reupload the downloaded image back to OMERO.

Result:
The image downloads as expected (it is a 3 image MIF, and the three images are downloaded locally).
Nevertheless, when the whole downloaded folder is reimported to OMERO, although the result are again 3 images, they do not appear to be multiimage file anymore, instead, for example, a Move into another group of only one of those images is possible. 
Imho, this is a test fail.

dominikl:
This also happens with the current Insight without this PR. Tbh, if download of the original files and reimport splits the multiimage Dicom image, than the problem is probably Dicom specific and lies deeper. Download and reimport of a vsi image for example works as expected with this PR.

pwalczysko:
Test 2:
1. Download a vsi, using the insight from the present PR, namely https://outreach.openmicroscopy.org/webclient/?show=image-94204
2. Reimport it again to OMERO as https://outreach.openmicroscopy.org/webclient/?show=image-94208

Result: The downloaded folder structure seems reasonable, see screenshot below.
When the vsi file is picked and reimported to OMERO, the viewing in OMERO is as expected, and it is a MIF as expected.
Screenshot 2021-12-13 at 16 45 40

This test passes imho.

pwalczysko:
> This also happens with the current Insight without this PR. Tbh, if download of the original files and reimport splits the multiimage Dicom image, than the problem is probably Dicom specific and lies deeper. Download and reimport of a vsi image for example works as expected with this PR.

Thank you @dominikl - confirming that with my release insight. The Dicom behaviour is not a regression, and has nothing to do with this PR.

joshmoore:
> If a file already exists, it will be skipped.

Would it be easy enough to expand this to "file _or directory_"? i.e. if *anything* already exists, you skip? I don't have a working example of how to trigger this, but I worried of downloading:

```
foo/my.ini
```
into a directory which already contains:
```
foo/my.fake
```
and thereby change the interpretation of something.

ehrenfeu:
Hi there,

quick and hopefully not so stupid question... Should the changes from above also be included in `omero-py` available from Jenkins? I'd be curious to test them.

Thanks,
Niko

pwalczysko:
> Hi there,
> 
> quick and hopefully not so stupid question... Should the changes from above also be included in `omero-py` available from Jenkins? I'd be curious to test them.
> 
> Thanks, Niko

Hi Niko,

The changes are certainly available on https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/ - just download the latest artifact of insight, that is what I am doing when testing. Not sure about omero-py, leave @dominikl to answer that.

dominikl:
As Petr said, the changes are included in the current merge build of Insight. But they won't go into omero-py, that's totally unrelated to Insight.

dominikl:
Last commit should address @joshmoore 's issue. Would be good to have some complex filesets (many files, many directories) to properly test this PR. What's a good image format for that?

pwalczysko:
> Last commit should address @joshmoore 's issue. Would be good to have some complex filesets (many files, many directories) to properly test this PR. What's a good image format for that?

Tested so far

- vsi
- ndpi
- svs
- Dicom (does not work, but not a regression)

The behaviour was as expected. @dominikl  to find one complex file format which will test it all is possibly not realistic ? Could we specify (with examples) what the "if the file already exists, it will be skipped" means, or, if that changed, how exactly and what is the expectation ? Only based on that we can set up some testing cases.

ehrenfeu:
> As Petr said, the changes are included in the current merge build of Insight. But they won't go into omero-py, that's totally unrelated to Insight.

Thanks Dominik. I somehow thought omero-py is calling the Java code when downloading files, but I should actually know better...

will-moore:
@ehrenfeu See https://forum.image.sc/t/download-full-projects-with-their-datasets/56434/22 and discussion there for downloading via python. Improvements were made in `omero-py 5.10.0`.

ehrenfeu:
Ah, very nice - thanks a lot @will-moore !

I was asking for a friend working on the HRM-OMERO connector 😁 but he already found out that he should just properly deal with the paths when downloading filesets from OMERO.

Nevertheless I will look into your code. My parts in the connector are *a bit* aged by now, they could certainly use some polishing.

sbesson:
Cross-linking to https://github.com/ome/omero-py/pull/298 which added the Python improvements that @will-moore was referring too. This thread in particular includes many considerations around data integrity and data corruption that are directly relevant to this use case.

Adding to the similarities, the former CLI download allowed the user to specify the name of the output file breaking the assumptions of the file readers when doing a round-trip. This is effectively another variant of https://github.com/ome/omero-insight/issues/188.

The agreement in OMERO.py was to 1- keep the structure of the original files (names & folder layout) unchanged, 2- enforce the containerisation of a fileset under a top-level folder. Talking with @pwalczysko @dominikl @will-moore and @khaledk2, I would strongly vote to use the similar strategy across our download utilities.

A difference between `omero download` and OMERO.insight download is that the latter supports the export of several filesets. In general, I think a export structure of type:

```
fileset1_folder/
   fileset1_originalfile_1
   fileset1_originalfile_2
   ...
   fileset1_originalfolder_1/
   ...
 fileset2_folder/
   fileset2_originalfile_1
   fileset2_originalfile_2
   ...
   fileset2_originalfolder_1/
   ...
```

is the one that is the most amenable to ensure the integrity of the original data and reduce the risks of cross-talk between filesets. In a nutshell, this is a small version of the ManagedRepository layout and matches the layout used in `omero-downloader`.

Probably the most important design decision for the layout above is the naming of these top-level fileset folders. There are a range of available options which offer trade-offs in terms of user-friendliness vs potential for interference with existing/other fileset folders.

joshmoore:
cc: @sukunis in case there are any cross-interactions with openlink.

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
```diff
@@ -3131,134 +3128,80 @@ IObject findIObject(SecurityContext ctx, String klassName, long id, boolean allG
...
-			while (i.hasNext()) {
-				set = (Fileset) i.next();
-				entries = set.copyUsedFiles();
-				j = entries.iterator();
-				while (j.hasNext()) {
-					FilesetEntry fs = j.next();
-					values.put(fs.getOriginalFile(), set);
+			for (Object tmp : files) {
+				Fileset fs = (Fileset) tmp;
+				File filesetDir = new File(dir.getAbsolutePath()+File.separator+"Fileset_"+fs.getId().getValue());
```
jburel, line 3139: method like ``Files.createDirectories`` could be used here instead.

pwalczysko:
Testing the behaviour of the repeated dowload of the same file into the same folder.
I have in Activities a message: 
``No image downloaded. Check logs for details.``
Then in the log I have 
``
2022-01-10 13:47:47,939 DEBUG [        o.o.shoola.env.data.OMEROGateway] ( Thread-46) /Users/pwalczysko/10Jan-test-download/Fileset_97307/Brain_DAPI_GFAP-488_IBA3-Cy3_MBP-647_coronal10x_coronal-slideA_03.vsi already exists. 
``
This makes sense and is as expected, but could we not have the main message saying ``..already exists`` in the Activities ? This would spare me having to go into the log - I have to know where the log is etc...

pwalczysko:
Tested so far
- vsi (also reimport)
- ndpi
- svs
- Dicom (does not work in the sense that the MIF is decoupled after reimport, but not a regression)

I think that the most logical to try now would be a plate download, but this si not enabled on merge-ci. What is the present workflow to change the config on that server please @sbesson ?

pwalczysko:
Downloading https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-4256 now. 


The https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-3855 is ending up being many single Filesets, possibly because it is a result of Dataset to Plate script ? 
Screenshot 2022-01-11 at 12 48 50

ehrenfeu:
In case you're interested, I have created a very small (~200 kB) multifile VSI dataset using the original Olympus cellSens software here at our facility. My purpose is to use it for automated testing of the HRM-OMERO connector, hence I tried to have it as small as possible while keeping the multifile structure (once it gets too small, cellSens puts everything into a single `.vsi` file without any subfolders).

Let me know if I should upload it somewhere!

src/main/java/org/openmicroscopy/shoola/env/data/OMEROGateway.java
dominikl: Doesn't really make it simpler, for `Files.createDirectory` I'd need to create a Path object instead of a File.

dominikl:
@pwalczysko Unfortunately I can only log the error if the file already exists. At the moment it's not possible to pass an error message up to the UI level. That would need a lot of changes to the Insight code.

pwalczysko:
The plates https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-4256 and https://merge-ci.openmicroscopy.org/web/webclient/?show=plate-2955  downloaded nicely, and the reimport as https://merge-ci.openmicroscopy.org/web/webclient/?show=screen-47354 went fine too. I do not see any problem there.

pwalczysko:
Any ideas for further testing formats @sbesson ?

Tested so far
 - vsi (also reimport)
 - ndpi
 - svs
 -  Dicom (does not work in the sense that the MIF is decoupled after reimport, but not a regression)
 - xdce extension (Plate)
 - Mias Frans (Plate, I think this is Flex ?)

joshmoore:
> https://github.com/ome/omero-insight/pull/264#issuecomment-1009932729 possibly because it is a result of Dataset to Plate script ?

That would make sense, @pwalczysko.

sbesson:
The list of formats tested in https://github.com/ome/omero-insight/pull/264#issuecomment-1011332008 should cover a wide-range of scenarios in different imaging domains (HCS, pathology). I assume NDPI included the NPDIS case (i.e. a pathology fileset composed of multiple NDPI files)?
https://docs.openmicroscopy.org/bio-formats/6.8.0/formats/dataset-table.html gives the full list of supported formats including the expectations in terms of dataset structure. The only other thing I can think of would be a multi-file format in the  fluorescence microscopy domain. Deltavision, Zeiss LSM and/or Olympus FV1000 might be good examples.

pwalczysko:
ndpi was https://merge-ci.openmicroscopy.org/web/webclient/?show=image-103615

pwalczysko:
> In case you're interested, I have created a very small (~200 kB) multifile VSI dataset using the original Olympus cellSens software here at our facility. My purpose is to use it for automated testing of the HRM-OMERO connector, hence I tried to have it as small as possible while keeping the multifile structure (once it gets too small, cellSens puts everything into a single `.vsi` file without any subfolders).
> 
> Let me know if I should upload it somewhere!

Hi Niko, 
Thank you very much for this. Would you be able to upload this file and make it publicly visible ? If you do not have a location, could you possibly use Zenodo ?
Thanks a lot
Petr

pwalczysko:
zeiss-lsm see https://merge-ci.openmicroscopy.org/web/webclient/?show=image-210326
deltavision with a log file https://merge-ci.openmicroscopy.org/web/webclient/?show=image-210329
fv1000 (oif with multi tiffs to it)  https://merge-ci.openmicroscopy.org/web/webclient/?show=image-210331

the three formats as above successfully downloaded and reimported. All MIFs and looked good after download and also after reimport.

This would be the coverage as suggested by comment https://github.com/ome/omero-insight/pull/264#issuecomment-1011930520

We have now


   - vsi (also reimport)
   - ndpi and ndpis
   - svs
   - Dicom (does not work in the sense that the MIF is decoupled after reimport, but not a regression)
   - xdce extension (Plate), with reimport
   - Mias Frans (Plate, I think this is Flex ?), with reimport
   - fv1000, reimported
   - Deltavision with log, reimported
   - zeiss-lsm (multifile version with mdb), also reimported
  
As all files and functional tests passed and the coverage seems rich enough, ready to merge fmpov.
Happy to add more tests if someone comes with more format ideas, otherwise all good.

sbesson:
Nothing else from my side. We might want to wait on the data from https://github.com/ome/omero-insight/pull/264#issuecomment-1012256622 to conduct a final test. Otherwise, I assume we are in a state where we should release OMERO.insight.

ehrenfeu:
Hi @pwalczysko

here you go: https://doi.org/10.5281/zenodo.5848987

Thanks a lot!
~Niko

pwalczysko:
@ehrenfeu Thanks a lot Niko !

Downloaded, imported to OMERO and again downloaded from OMERO using insight from this PR.
All went fine, the set consists of 2 images in OMERO, a 2-channel image with 11 z sections and other parameters matching the README of your file in Zenodo. The second image is Macro image. See screenshot below documenting  how the hierarchy looks like on my local Mac after download and how the images look in OMERO.iviewer after reimport. I cannot detect any problem there, all works and matches the images pre-download.

Screenshot 2022-01-14 at 12 01 27

Screenshot 2022-01-14 at 12 04 11

joshmoore:
cc: @erickmartins since he may have something similar in his upcoming transfer tools

https://twitter.com/erickratamero/status/1482060699539058688

erickmartins:
We live in a python/ezomero world, so our strategy is to use our [`get_original_filepaths`](https://github.com/TheJacksonLaboratory/ezomero/blob/b6650a7a30ca11f01e7a5e669927573b3d554f1c/ezomero/_gets.py#L876) ezomero function (which HQLs for `usedFiles` in a Fileset) and `rsync` via `subprocess` maintaining the internal `ManagedRepository` directory structure. In our experience it has dealt with all cases pretty cleanly. If fileset downloading is included in `omero-py` we could potentially lose the rsync dependency and do it that way.

(we're probably making our transfer thing public this week. the repo is still private because we didn't want a half-half-baked solution in the hands of people, and because we still need a catchier name than `jax-omero-transfer` :) )

pwalczysko (approved):

ehrenfeu:
Thanks J-M, highly appreciated!

jburel:
@ehrenfeu https://github.com/ome/omero-insight/releases/tag/v5.7.0 containing this PR is now available

PR #266, opened 2022-01-05
fix NPE if no user is displayed

jburel:
Fix https://www.openmicroscopy.org/qa2/qa/feedback/30781/

cc @dominikl


dominikl:
👍 Thanks, looks good.

PR #267, opened 2022-01-05
handle possible null object

jburel:
Fix for https://www.openmicroscopy.org/qa2/qa/feedback/30778/ 
cc @dominikl


dominikl:
Looks good, should avoid the NPE 👍

PR #270, opened 2022-01-12
QA 30792

jburel:
See https://www.openmicroscopy.org/qa2/qa/feedback/30792/
handle path > 1
cc @dominikl


dominikl:
Do yo know how this can be replicated? Something like an OMERO script with an empty String as a path (what kind of path is that)?

dominikl:
Ah I see, that's also mentioned on https://forum.image.sc/t/error-in-insight-after-uploading-script/61807 .

dominikl:
Good to merge. Don't know how to replicate the issue, but the change makes sense, and script functionality still works as expected 👍

Issue #271, opened 2022-01-14
Download popup "File format" chooser


pwalczysko:
Download of original file popup shows a dysfunctional dropdown menu saying File format.

Workflow:
1. Start insight
2. Connect to server,
3. Select any image which has original file to it (not inside Plaate, as this might not have the Download enabled on your server)
4. Click onto the Download dropdown in top-right, select the first option, with blue downward facing arrow to dowload the original file
5. Observe that in the bottom of the first dialog (the download location chooser), there is the reduntant "File format" dropdown

![Screenshot 2022-01-14 at 11 45 18](https://user-images.githubusercontent.com/2478303/149510739-ad9076ed-5e80-47d5-af09-00685a3713f6.png)

cc @dominikl @jburel

dominikl:
👍 Thanks Petr. I'll open a separate PR for that, to keep it apart from the current open "download" PR https://github.com/ome/omero-insight/pull/264 .

PR #272, opened 2022-01-14
Disable file format combobox

dominikl:
Workaround for https://github.com/ome/omero-insight/issues/271 . This "File Format" dropdown is part of Java's builtin JFileChooser, so I can't remove it. This PR is a proposed workaround which simply disables it in case it's empty.
Screenshot 2022-01-14 at 13 53 39

Test: Make sure it's still enabled when it should be (e.g. attach a file), and that it is disabled when it is empty (e.g. download an image).

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/util/ui/filechooser/CustomizedFileChooser.java
+++ b/src/main/java/org/openmicroscopy/shoola/util/ui/filechooser/CustomizedFileChooser.java
@@ -136,13 +136,13 @@ class CustomizedFileChooser
 					label.setText(FOLDER_LABEL);
 				setFileSelectionMode(DIRECTORIES_ONLY);
 				setCurrentDirectory(getFileSystemView().getHomeDirectory());
-				// Workaround to disable the (empty) file formats dropdown box:
-				List<Component> comps = UIUtilities.findComponents(this, JComboBox.class);
-				for (Component comp : comps) {
-					JComboBox box = (JComboBox)comp;
-					if (box.getItemCount() <= 1)
-						box.setEnabled(false);
-				}
+		}
+		// Workaround to disable useless file formats dropdown box:
+		List<Component> comps = UIUtilities.findComponents(this, JComboBox.class);
+		for (Component comp : comps) {
+			JComboBox box = (JComboBox)comp;
+			if (box.getItemCount() <= 1)
+				box.setEnabled(false);
 		}
 	}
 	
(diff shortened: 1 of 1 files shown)


pwalczysko:
Thanks @dominikl - not sure how much of a better experience is that though. The dialog is still there... It is just that I cannot click on the chooser, which is empty. Checking the other workflows, note that when downloading an attached file, the box is active, and allways shows "All files" - I can even actively choose, but only one option only :)
Screenshot 2022-01-19 at 17 29 46


Is this improvement really worth any risks of regression ?

pwalczysko:
Also checked the export as png, jpeg, tiff and ome-tiff. Cannot see anything wrong with any of the options, the first three are exporting the first plane, with all channels separate and a merge. The ome-tiff option gives me for example all the timepoints with one image with correct number of channels which can be successfully reimported into OMERO and looks good.

But for the improvement iintroduced in this PR, I am really not sure -  the confusing dialog is still present, see https://github.com/ome/omero-insight/pull/272#issuecomment-1016701962  What is your opinion, @jburel ?

dominikl:
Thanks Petr. I'll add another commit and remove the "All Files" option for that case (that should then disable the drop down as well).

dominikl:
I just disabled the file type drop down if there isn't any choice (ie. less than 2 options).

jburel:
Note JFilechooser implementation is different between OS. This should be checked minimally on Mac and Windows

pwalczysko:
> Note JFilechooser implementation is different between OS. This should be checked minimally on Mac and Windows

Tested on Windows and Mac. All works as described.
Still not sure that is the way other apps are doing it, we are well off the common path here imho. Studying PowerPoint and Excel, the Format chooser is either fully functional (on export as, with many options to choose) or is not there at all (on Open commands in both applications). When the Open/Upload/Import command is used (in Excel there is an Import command), then the main browser box again has no Format filter choice, the choice is done in a dedicated cialog beforehand.
I think it is fine to merge, but just not sure about the benefit.

dominikl:
Merge @jburel ? I still think it's better to disable a pointless drop down box as we can't remove it.

jburel:
I will have a look at the code change.

jburel:
@pwalczysko the problem is that we are using a built-in component ``JFileChooser`` from swing. Comparing to Excel or Powerpoint is not really a suitable way to check.
Disabling it is fine. I don't think we should spend too much time fighting with that the `JFileChooser`` component which has historically been problematic.

PR #276, opened 2022-01-31
QA 30817

jburel:
handle case when the browser is null


dominikl:
Not sure how to replicate the issue, but the change looks good. Download functionality also still works like expected 👍

Issue #274, opened 2022-01-24
IllegalComponentStateException: TreeViewerWin.showPopup()


dominikl:
[QA 30811](https://www.openmicroscopy.org/qa2/qa/feedback/30811/)

```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.awt.IllegalComponentStateException: component must be showing on the screen to determine its location
	at java.awt.Component.getLocationOnScreen_NoTreeLock(Component.java:2062)
	at java.awt.Component.getLocationOnScreen(Component.java:2036)
	at javax.swing.JPopupMenu.show(JPopupMenu.java:948)
	at org.openmicroscopy.shoola.agents.treeviewer.view.TreeViewerWin.showPopup(TreeViewerWin.java:786)
...
```

jburel:
see also https://www.openmicroscopy.org/qa2/qa/feedback/30824/

dominikl:
Fixed by https://github.com/ome/omero-insight/pull/290 .

PR #290, opened 2022-04-11
Make sure component is showing

dominikl:
I don't know how to replicate https://github.com/ome/omero-insight/issues/274 but this PR should prevent it.


jburel:
Code change makes sense.
Did you review other occurrences of dialog popping up?

dominikl:
There are actually a lot of locations where `JPopupMenu.show(component, ...)` is called (~30 places), but none except this one seems to be a problem.

jburel:
Thanks for checking

PR #295, opened 2022-05-03
Issue 210: javax.swing.JComponent: ArrayIndexOutOfBoundsException

dominikl:
Another attempt to fix https://github.com/ome/omero-insight/issues/210 .

Used the `CheckThreadViolationRepaintManager` mentioned in https://stackoverflow.com/questions/3153574/how-to-detect-swing-thread-policy-violations to detect more code in the import context which triggers UI updates outside the Swing thread. Moved these parts into `SwingUtilities.invokeLater()` calls. 

**Testing**: The exceptions of issue 210 are random and not reproducable. But what needs testing is that the import workflows still work.

/cc @jburel

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/Main.java
+++ b/src/main/java/org/openmicroscopy/shoola/Main.java
@@ -81,7 +81,7 @@ public class Main
 		if (posArgs.size() > 1) homeDir = posArgs.get(1);
 
 		try {
-			Path p = Path.of(Container.CONFIG_DIR, Container.CONFIG_FILE);
+			Path p = Path.get(Container.CONFIG_DIR, Container.CONFIG_FILE);
 			String content = Files.readString(p).replaceAll("\n", "");
 			Matcher m = Pattern.compile("DebugRepaintManager.+?>(.+?)<").matcher(content);
 			if (m.find() && Boolean.parseBoolean(m.group(1))) {
--- a/src/main/java/org/openmicroscopy/shoola/Main.java
+++ b/src/main/java/org/openmicroscopy/shoola/Main.java
@@ -29,6 +29,7 @@ package org.openmicroscopy.shoola;
 
 //Application-internal dependencies
 import org.openmicroscopy.shoola.env.Container;
+//import org.openmicroscopy.shoola.util.CheckThreadViolationRepaintManager;
 
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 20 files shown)


src/main/java/org/openmicroscopy/shoola/Main.java
```diff
@@ -72,6 +73,10 @@ public static void main(String[] args)
 		List<String> posArgs = Arrays.stream(args).filter(a -> !a.startsWith("--")).collect(Collectors.toList());
 		if (posArgs.size() > 0) configFile = posArgs.get(0);
 		if (posArgs.size() > 1) homeDir = posArgs.get(1);
+
+		// For debugging Swing thread exceptions:
+		//RepaintManager.setCurrentManager(new CheckThreadViolationRepaintManager());
```
joshmoore, line 78: Nice find! Not sure if we have a "debug" setting, but would this be something it would be useful for others to run, or is more dev only?

dominikl: Creds to @jburel , he pointed me to the stackoverflow post. I thought I add the `CheckThreadViolationRepaintManager` class, might be useful in future for similar issues. But yes, maybe there's a better way to do it than just commenting it out.

jburel: An option could be to add a flag in ``container.xml`` so we can turn the option on/off

dominikl: Ta, I'll do that.

src/main/java/org/openmicroscopy/shoola/Main.java
```diff
@@ -74,8 +80,16 @@ public static void main(String[] args)
 		if (posArgs.size() > 0) configFile = posArgs.get(0);
 		if (posArgs.size() > 1) homeDir = posArgs.get(1);
 
-		// For debugging Swing thread exceptions:
-		//RepaintManager.setCurrentManager(new CheckThreadViolationRepaintManager());
+		try {
+			Path p = Path.of(Container.CONFIG_DIR, Container.CONFIG_FILE);
```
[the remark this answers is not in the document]
dominikl: `Path.get` doesn't seem to exist in Java 11. 
``` 
error: cannot find symbol
Path p = Path.get(Container.CONFIG_DIR, Container.CONFIG_FILE);
			^
symbol:   method get(String,String)
```

jburel:
Looking at the code, the ``JXBusyLabel`` could be the source of the issue.
Other usage of the component could potential trigger the error  e.g.  when checking the disk space available 
 (``UserDiskSpace``), search for data (``SearchComponent``)

dominikl:
I wrapped the `setStatus` calls to JXBusyLabel into `SwingUtilities.invokeLater()`; but only "update" calls, not if just used for initializing a UI component.

dominikl:
Could this be merged now? I'll definitely get merge conflicts with subsequent PRs, but I think this one would be best to get in first.

jburel:
Sorry I thought I merged it

PR #296, opened 2022-05-13
Remove joinSession

dominikl:
Remove the calls to joinSession, see https://github.com/ome/omero-gateway-java/pull/63 .


dominikl:
Could we merge this PR together with https://github.com/ome/omero-gateway-java/pull/63 @jburel ?
Only change for the user will be that he has to paste the sessionID as username and as password, not only as username. But that would match with the python api/omero.web.

jburel:
I don't think that behaviour is documented anywhere. I know that ssbd uses the session to connect to the server. That will need to synch with them

dominikl:
Added a workaround if the session id is only pasted as username and password is empty. So no changes in behaviour for users.

PR #297, opened 2022-05-18
handle case without tab

jburel:
With an error reported via https://www.openmicroscopy.org/qa2/qa/feedback/31082/
Code review is enough


dominikl:
I think the issue was that one could close the last and only Importer tab, then you got this error. Should have been fixed a while ago with another PR, preventing that. But it doesn't do any harm to have another check and return if the index is < 0. 👍

dominikl (approved):

PR #298, opened 2022-05-20
Bug: NPE if user only in system group

jburel:
To reproduce the issue, use a user only in the system group


dominikl (approved):
Looks good to me 👍

PR #299, opened 2022-05-20
Bug: No image to save

jburel:
handle case when the main image does not exist e.g. cannot be loaded


dominikl (approved):
Looks good to me 👍

PR #302, opened 2022-05-24
Hide groups if session key is used to connect

jburel:
Do not show the groups if a session key is used to connect

To test:
* Connect using username and password using a user in multiple groups. Check that the user can access the groups via the UI.
* Connect using a session key. Check that the user cannot access the other groups.


dominikl:
👍  Looks good. Works like expected.

dominikl (approved):

Issue #308, opened 2022-06-06
Loading ROIS


jburel:
loading large number of rois could lead to exception like the one reported in https://www.openmicroscopy.org/qa2/qa/feedback/31112/
cc @dominikl

PR #318, opened 2022-08-17
Don't load annotations for ROIs

dominikl:
Just removes the call to load annotations for ROIs (we don't do anything with them anyway).
Should fix https://github.com/ome/omero-insight/issues/308 .


jburel:
Changes make sense.
Do you remember why we added the call if we did not display the annotations?

dominikl:
No, I was wondering too. There's probably more potential code to clean up/remove with that respect. Is it worth doing within this PR?

jburel:
If we are not displaying anything to the users, I don't think we should spend more time cleaning the code.

dominikl:
I won't do any more old code clean up then. Good to merge @jburel ?

Issue #325, opened 2022-11-10
InvalidPathException: Malformed input or input contains unmappable characters


dominikl:
[QA 31322](https://www.openmicroscopy.org/qa2/qa/feedback/31322)
Path name with `?` in it crashes Insight (althought I think this might actually be some unicode characters etc).

jburel:
Fixed https://github.com/ome/omero-insight/pull/328

Issue #174, opened 2020-11-05
File chooser intermittent error


jburel:
```
ava.nio.file.InvalidPathException: Illegal char <:> at index 21: C:\OMERO.insight\app\::{F02C1A0D-BE21-4350-88B0-7367FC96EF3C}


	at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)


	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)


	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)


	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
```

see https://www.openmicroscopy.org/qa2/qa/feedback/29653/

dominikl:
There are two other very similar issues: #111 and  #146 . And I still don't know how to replicate either of them.

dominikl:
[QA 29996](https://www.openmicroscopy.org/qa2/qa/feedback/29996/)

dominikl:
[QA 30308](https://www.openmicroscopy.org/qa2/qa/feedback/30308/)

dominikl:
[QA 30806](https://www.openmicroscopy.org/qa2/qa/feedback/30806/)

dominikl:
[QA 31223])https://www.openmicroscopy.org/qa2/qa/feedback/31223/

jburel:
Fixed https://github.com/ome/omero-insight/pull/328 (version 5.8.0)

PR #328, opened 2022-11-24
check if the path is valid

jburel:
This PR handles the attempt to import with invalid characters in the path.
The files will no be added to the list of files to import


dominikl:
Looks good, thanks @jburel 👍 Fixes https://github.com/ome/omero-insight/issues/325 , https://github.com/ome/omero-insight/issues/174 and https://github.com/ome/omero-insight/issues/146 . Haven't tested it though. Is there a way to test it? I struggled to replicate these errors in the first place.

dominikl (approved):

PR #329, opened 2022-11-25
DataServicesFactory: only log server exception class and message

sbesson:
Similar to d17677fa6 this avoids printing the entire server stack trace to the log file is the client is talking to an older server where the property is not exposed client-side.


jburel (approved):
Changes match what has already been implemented when retrieving another property

Issue #343, opened 2023-02-06
NPE: ThumbnailLoader


dominikl:
[QA 32573](https://www.openmicroscopy.org/qa2/qa/feedback/32573/)

"Selected a dicom image in orphaned images."

```
java.lang.Exception: java.lang.NullPointerException
	at org.openmicroscopy.shoola.env.data.views.calls.ThumbnailLoader$1.doCall(ThumbnailLoader.java:232)
```

PR #344, opened 2023-02-06
Avoid NPE

dominikl:
That should "fix" issue https://github.com/ome/omero-insight/issues/343 . Tbh, I don't know how to replicate the issue, nor how it's possible that the pixels or image object is null at that point, but the PR should "fix" the problem by just skipping over this image.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
@@ -215,8 +215,10 @@ public class ThumbnailLoader extends BatchCallTree {
             int k = 0;
             for (DataObject image : images) {
                 // Cast our image to pixels object
+                if (image == null)
+                    continue;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ThumbnailLoader.java
```diff
@@ -216,6 +216,8 @@ protected void buildTree() {
             for (DataObject image : images) {
                 // Cast our image to pixels object
                 final PixelsData pxd = dataObjectToPixelsData(image);
```
jburel, line 218: ``image == null`` should be checked before invoking ``dataObjectToPixelsData``

jburel: This could happen if an image has no pixel.

dominikl: 👍

PR #348, opened 2023-02-17
Add changelog entry for 5.8.1 release

dominikl:
There seems to be daily QA with the scripts menu issue. Maybe we should do a wee bugfix release @jburel ?
Could get these three PRs in for that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+5.8.1 (February 2023)
+---------------------
+
+- Fix issue with scripts menu (https://github.com/ome/omero-insight/pull/341)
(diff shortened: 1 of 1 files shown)


jburel:
@dominikl The failure is due to the fact that ``Gradle 8`` is used during the evaluation phase. The builds that are green were using ``7.6``.
We had a similar problem with gradle 7 previously

jburel:
https://github.com/ome/omero-insight/pull/349 fixes the issue

jburel:
@dominikl I have added a few more entries to the changelog.

dominikl:
👍 Looks good.

PR #220, opened 2021-04-20
Gradle7 fix

jburel:
This PR tests the build in GHA with Gradle 6.8.3
The changes should work with Gradle 7. 
The failure happened when running ``gradle wrapper --gradle-version=6.8.3``
cc @dominikl

PR #349, opened 2023-02-21
use gradle action

jburel:
Using the Gradle action installed the desired version of gradle preventing the failure. see https://github.com/ome/omero-insight/actions/runs/4202376704/jobs/7353445683
TODO: review other repositories

Check that the build is green

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -16,7 +16,7 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
     runs-on: ${{ matrix.os }}
     env:
-      gradle_version: 6.8.3
+      gradle_version: 6.8.3 # use ``current`` to install the current stable release
       gradle_commands: build # default is build
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


dominikl:
👍 Looks good. These github actions seem to be quite a pain to maintain...

jburel:
It is not really a pain, things evolved as the quality of the tooling increases.
the gradle-action used her was not around (or I was not aware of it) when we first started using them

sbesson:
As we might be looking into upgrading Gradle to 7.x across the board, is there a track record of the rationale and the initial decision not to use the Gradle Wrapper across the new Java repositories?

Obviously there are advantages and drawbacks to either approach. In our case, the Jenkins and the super-module infrastructure are likely implemented without this assumption and would require modification. One advantage from a developer perspective is that it declares the Gradle version in an explicit manner. At the CI level, I assume this would remove the requirement to manage the Gradle version both in the GitHub workflow and within the Jenkins node, reducing the need for PRs such as this one.

jburel:
Other repositories use https://github.com/ome/action-gradle/pull/2 
Using the wrapper is also an option with that action.
But as we have noticed before 
it can lead to error during the "installation" of the wrapper see for example https://github.com/ome/omero-insight/pull/220

This is exactly what happens here with Gradle 8, some "check-up" tasks are run with the current release (gradle 8) before actually using the wrapper.
The wrapper does not isolate unfortunately

jburel:
We will revisit the strategy post release i.e. include or not gradlew file.
Merging so we can build the release

Issue #363, opened 2023-06-15
MacOS: edit or store container.xml outside the application to avoid SystemPolicyAppBundles restrictions


TK5-Tim:
Is there a way to store the container.xml for macOS outside the application to prevent any issues with [SystemPolicyAppBundles](https://macmule.com/2022/10/24/macos-ventura-and-bypassing-the-new-systempolicyappbundles-privacy-policy-control/)? 

If not that would be very helpful to prevent issues in the future otherwise it could happen that the application fails to work after changing the container.xml file. 

Thanks
Tim Keller

dominikl:
Hi Tim,

I'm not quite sure what the problem is. If I edit something within the OMERO.insight.app, is that then supposed to trigger some kind of security error? On my OSX (12.6.5) it doesn't. 

In the OMERO.insight.cfg you can set 
```
[ArgOptions]
container.xml
```
to something else, e.g. containerImporter.xml to load the Importer UI, but you can't point it to a file outside the app package. Is it that what you meant, that this should be possible?

TK5-Tim:
Yes, currently it is not triggering the security error, because the application is not signed. 
Because there is more and more security restrictions by Apple that could change in the future. 

For signed applications that is already the case with macOS 13.0+ 
In the future that could be also the case with unsigned applications. 

Pointing it to a file outside the app package would be helpful and one way to avoid that problem.

PR #365, opened 2023-06-16
Enable pointing to external container xml

dominikl:
Fixes https://github.com/ome/omero-insight/issues/363 .

You can start Insight with a positional argument pointing to the container xml (e.g. container.xml for the "normal" Insight, containerImport.xml for the Importer). In the OSX build this argument is set by OMERO.insight.cfg:
```
[ArgOptions]
container.xml
```

Without this PR the container xml has to be in the config directory within the bundle. With this PR you can point to any location. If the argument contains at least one path character it is assumed it specifies a file outside the bundle, if it doesn't (i.e. it's only a file name) it's assumed that it specifies a file within config directory of the bundle.

The PR also removes the code to enable the DebugRepaintManager, as it's actually not used any more.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -87,3 +87,15 @@ You must edit the config file `app\OMERO.insight.cfg` stored in application fold
  - Unzip the app.
  - Open the file ``bin/omero-insight`` in a text editor.
  - Change the value of ``DEFAULT_JVM_OPTS``.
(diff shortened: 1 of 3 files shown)


jburel:
Instructions should be added to the FAQ indicating how to use external config file

PR #382, opened 2023-08-05
adjust title in ImageJ/Fiji window

jburel:
adjust title in ij window: title is used to set image name
See https://github.com/ome/omero-insight/issues/362 for testing info


dominikl (approved):
Looks good 👍

PR #384, opened 2023-08-07
replace : by _ in name on Windows

jburel:
Replace special character by underscore in the name to avoid exception on Windows
see #/347


dominikl (approved):
Makes sense 👍

Issue #360, opened 2023-05-09
NPE: table


jburel:
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.IllegalArgumentException: Row index out of range
	at javax.swing.JTable.boundRow(JTable.java:2146)
	at javax.swing.JTable.addRowSelectionInterval(JTable.java:2196)
	at com.apple.laf.AquaFileChooserUI$DnDHandler$1.run(AquaFileChooserUI.java:795)
	at java.awt.event.InvocationEvent.dispatch(InvocationEvent.java:311)
	at java.awt.EventQueue.dispatchEventImpl(EventQueue.java:758)
	at java.awt.EventQueue.access$500(EventQueue.java:97)

```
see https://www.openmicroscopy.org/qa2/qa/feedback/32939/

PR #385, opened 2023-08-07
handle case where ROI is not longer in table

jburel:
see #360

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/measurement/view/ServerROITable.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/measurement/view/ServerROITable.java
@@ -313,13 +313,20 @@ class ServerROITable
 				l.add(array[j]);
 		}
 		if (roiIDs.size() <= l.size()) table.clearSelection();
+		// Check the number of rows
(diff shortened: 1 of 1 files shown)


dominikl:
Is there a way to replicate the error?

src/main/java/org/openmicroscopy/shoola/agents/measurement/view/ServerROITable.java
```diff
@@ -313,13 +313,20 @@ void selectROI(List<Long> roiIDs)
...
 			if (rowIDs.containsKey(id)) {
 				index = rowIDs.get(id);
-				if (!l.contains(index)) {
-					table.addRowSelectionInterval(index, index);
-					scrollToRow(index);
+				if (index >= count) {
+					table.clearSelection();
+					break;
+				} else {
+                    if (!l.contains(index)) {
```
dominikl, line 326: The indentation is a bit off here.

jburel: probably because we have a mix of space and tab in the file

Issue #361, opened 2023-05-17
Open image on Fiji when using the search bar


Rdornier:
## Enhancement report

**Description of the expectation**
When you search for an image using the search bar, there is no easy way to open it on Fiji. Double-clicking on the image doesn't open it on Fiji. Moreover, There is no _browse_ button to directly get into the hierarchy and highlight the image (like in the OMERO.web browser).

This is particularly annoying when we have only the image ID. The search bar find quickly the image but then, we have to manully get the parent dataset/project, copy the name of the image somewhere, browse in the hierarchy the image and finally double-click on it to open in Fiji.

It would be very cool having an easy way to open an image on Fiji when using the search bar, like the _browse_ button available.

![image](https://github.com/imagej/imagej-omero/assets/64911638/2787457e-aeb8-4f1e-a281-7e3055eb8061)

**Plugin**
omero_ij-5.81-all.jar

@romainGuiet

PR #386, opened 2023-08-07
open image in imageJ when using as IJ plugin

jburel:
see https://github.com/ome/omero-insight/issues/361

The View button displayed in the table will open the image in Fiji when used in the plugin
To test: 
* search for an image e.g. using id
* in the result table. Click on the ``View`` button.
* The image should open in Fiji


dominikl (approved):
👍

Issue #358, opened 2023-04-11
NPE DownloadActivity.getFileName


dominikl:
[QA 32874](https://www.openmicroscopy.org/qa2/qa/feedback/32874/)

```
java.lang.NullPointerException
	at org.openmicroscopy.shoola.env.ui.DownloadActivity.getFileName(DownloadActivity.java:112)
	at org.openmicroscopy.shoola.env.ui.DownloadActivity.(DownloadActivity.java:190)
...
```
"...the crash occurred only if I tried to download file directly to some root directory (aka c:\\, or d:\\), download to any (sub)folder went fine."

PR #387, opened 2023-08-07
Handle root directory

jburel:
This PR should handle the case when attempting to download files in root directory
see #358


dominikl (approved):
👍

Issue #380, opened 2023-07-26
CCE: DatasetData cannot be cast to TreeImageSet


dominikl:
[QA 33723](https://www.openmicroscopy.org/qa2/qa/feedback/33723)
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.ClassCastException: omero.gateway.model.DatasetData cannot be cast to org.openmicroscopy.shoola.agents.util.browser.TreeImageSet
org.openmicroscopy.shoola.agents.treeviewer.RefreshExperimenterDataLoader.formatSmartFolderResult(RefreshExperimenterDataLoader.java:199)
	at org.openmicroscopy.shoola.agents.treeviewer.RefreshExperimenterDataLoader.handleResult(RefreshExperimenterDataLoader.java:382)
...
```

dominikl:
Additional info from user:
"Running Cell Profiler + AI
I think I was trying to create an Image List (about 40 images).
It might have timed out."

jburel:
See https://github.com/ome/omero-insight/pull/389

PR #389, opened 2023-08-07
Pass the node and not the Data object

jburel:
Pass the node and not the Data object see #380

Check that you can refresh.


dominikl:
When does this error happen, refresh what?

jburel:
You need to refresh the various views via the left-hand harmonicas

dominikl (approved):
Still can't replicate the error. But I guess it's fine 👍

Issue #367, opened 2023-06-26
NPE: create object 


jburel:
```
java.lang.Exception: java.lang.IllegalArgumentException: IObject cannot be null.
	at omero.gateway.util.PojoMapper.asDataObject(PojoMapper.java:165)
	at org.openmicroscopy.shoola.env.data.OmeroDataServiceImpl.createDataObject(OmeroDataServiceImpl.java:346)
	at org.openmicroscopy.shoola.env.data.views.calls.DataObjectSaver$1.doCall(DataObjectSaver.java:79)

....
       at org.openmicroscopy.shoola.agents.treeviewer.DataTreeViewerLoader.handleException
```

QA https://www.openmicroscopy.org/qa2/qa/feedback/33433/

jburel:
see #391

PR #391, opened 2023-08-07
handle Null object when attempting to create an object

jburel:
The easiest way to test this PR
is to return ``null`` instead of the ``pojo`` at https://github.com/ome/omero-insight/blob/master/src/main/java/org/openmicroscopy/shoola/env/data/OmeroDataServiceImpl.java#L346 (now line 348)

see #367


dominikl (approved):
👍 Good idea, thanks JM.

Issue #364, opened 2023-06-15
ArrayIndexOutOfBoundsException FailedImportDialog


dominikl:
[QA 33046](https://www.openmicroscopy.org/qa2/qa/feedback/33046/)
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.lang.ArrayIndexOutOfBoundsException: 0
	at org.openmicroscopy.shoola.agents.fsimporter.view.FailedImportDialog.buildUI(FailedImportDialog.java:110)
	at org.openmicroscopy.shoola.agents.fsimporter.view.FailedImportDialog.(FailedImportDialog.java:80)
	at org.openmicroscopy.shoola.agents.fsimporter.view.ImporterUIElementLight.showFailures(ImporterUIElementLight.java:276)
...
```

jburel:
See also https://www.openmicroscopy.org/qa2/qa/feedback/33052/

PR #392, opened 2023-08-13
display failure dialog if any

jburel:
Code review
This should avoid https://github.com/ome/omero-insight/issues/364


dominikl (approved):
👍

PR #354, opened 2023-03-20
Close all HandleTie servants created during import

sbesson:
This addresses an issue with associated with dangling servants left at the end of a data import. This PR copies over the minimal callback closing changes from the `ImportLibrary.importImage` and more specifically https://github.com/ome/omero-blitz/blob/49c5d46042487014d0dd6753679cf7f15bee873b/src/main/java/ome/formats/importer/ImportLibrary.java#L704-L709 to ensure that all handle servants are unregistered server-side.

To test these changes, run an import of one or multiple filesets and watch the creation/deletion of `HandleTie` servants server-side e.g. using `tail -F Blitz-0.log  | grep HandleTie`.

- without this PR, the import should create N `omero.cmd._HandleTie` servants, one per fileset, and only close N-1. The last servant should be created (`Added servant...`) but never deleted (`Unregistered servant...`)
- with this PR included, N servants should be created and successfully deleted

PR #394, opened 2023-08-25
Close all HandleTie servants created during import

jburel:
This PR replaces #354

Run an import of one or multiple filesets and watch the creation/deletion of HandleTie servants server-side e.g. using ``tail -F Blitz-0.log  | grep HandleTie.``
All the handles should now be closed and the service closed.
Thumbnails are no longer loaded

To test:
* regular import of single files
* regular import of directories (1-10 & >50 to test the lightfileimport)
* import of broken files (failing server-side)
* import of HCS data
* import of some images which take longer to process on the server side and closing insight before the server side import is finished.

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
@@ -405,6 +405,11 @@ public class FileImportComponent
 	/** Indicates that the import was successful or if it failed.*/
 	private void formatResult()
 	{
+		if (callback != null) {
+			try {
+				((CmdCallbackI) callback).close(true);
+			} catch (Exception e) {}
+		}
 		SwingUtilities.invokeLater(() -> {
 			if (namePane.getPreferredSize().width > LENGTH)
 				fileNameLabel.setText(EditorUtil.getPartialName(
@@ -897,7 +902,7 @@ public class FileImportComponent
 					formatResultTooltip();
 				}
 				else if (image instanceof Collection){
-					Collection<?> c = (Collection)image;
+					Collection<?> c = (Collection) image;
 					if(!c.isEmpty()) {
 						Object obj = c.iterator().next();
 						if(obj instanceof ThumbnailData) {
@@ -966,11 +971,6 @@ public class FileImportComponent
 						FileImportComponent.this.image = null;
 					}
 				}
-				if (callback != null) {
-			        try {
-				        ((CmdCallbackI) callback).close(true);
-			        } catch (Exception e) {}
-		        }
 				FileImportComponent.this.invalidate();
 			}
 		});
@@ -1581,8 +1581,14 @@ public class FileImportComponent
 		});
 	}
 	
+	/**
+	 * Sets the status depending on outcome
+	 */
 	private void setImportResult() {
 		Object result = status.getImportResult();
+		if (result == null) {
+			resultIndex = ImportStatus.FAILURE;
+		}
 		if (image instanceof ImportException) result = image;
 		if (result instanceof ImportException) {
 		    ImportException e = (ImportException) result;
@@ -1593,9 +1599,7 @@ public class FileImportComponent
 			else if (status == ImportException.MISSING_LIBRARY)
 				resultIndex = ImportStatus.FAILURE_LIBRARY;
 			else resultIndex = ImportStatus.FAILURE;
-		} else if (result instanceof CmdCallback) {
-			callback = (CmdCallback) result;
-		} else {
+		} else if (result instanceof Collection) {
 			resultIndex = ImportStatus.SUCCESS;
 		}
 	}
--- a/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
+++ b/src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
@@ -405,11 +405,6 @@ public class FileImportComponent
 	/** Indicates that the import was successful or if it failed.*/
 	private void formatResult()
 	{
-		if (callback != null) {
-			try {
-				((CmdCallbackI) callback).close(true);
-			} catch (Exception e) {}
-		}
 		SwingUtilities.invokeLater(() -> {
 			if (namePane.getPreferredSize().width > LENGTH)
 				fileNameLabel.setText(EditorUtil.getPartialName(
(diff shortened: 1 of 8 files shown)


src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
```diff
@@ -1591,6 +1581,29 @@ public void propertyChange(PropertyChangeEvent evt)
...
+	 */
+	private void setImportResult() {
+		Object result = status.getImportResult();
+		if (result == null) {
+			resultIndex = ImportStatus.FAILURE;
+		}
+		if (image instanceof ImportException) result = image;
+		if (result instanceof ImportException) {
+		    ImportException e = (ImportException) result;
+		    image = result;
```
sbesson, line 1595: This call was not present in the original logic of `formatResult` and is not present in `LightFileImportComponent`. What is its function?

src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ImagesImporter.java
```diff
@@ -124,7 +124,7 @@ protected void buildTree()
             final boolean b = index == n;
             index++;
             add(new BatchCall("Importing file") {
-                public void doCall() { importFile(f, b); }
+                public void doCall() { importFile(f, false); }
```
sbesson, line 127: Should `final boolean b = index == n;` be removed if it's no longer used?

src/main/java/org/openmicroscopy/shoola/agents/fsimporter/view/ImporterModel.java
```diff
@@ -356,6 +357,19 @@ void fireImportData(ImportableObject data, int loaderID)
...
+		ImagesImporter loader = loaders.get(loaderID);
+		if (loader != null) {
+			ImportableObject io = loader.getImportableObject();
+			try {
+				ImporterAgent.getRegistry().getImageService().closeImport(io);
+			} catch (Exception e) {
+				LogMessage msg = new LogMessage();
+                msg.print("Import closure");
+                msg.print(e);
+				ImporterAgent.getRegistry().getLogger().info(this, msg);
```
sbesson, line 369: Should this be a `warn` instead? I assume in most cases we are exception `closeImport` to complete without exceptions?

sbesson (changes requested):
Tested against a remote OMERO server in a separate geographical region using a local build of OMERO.insight from the HEAD of this PR across the scenarios described in the description, specifically:
- two folders containing each 10 & 30 fake files
- a `plate.fake` file with an accompanying `plate.fake.ini` with `plateRows=4`, `plateCols=4` and `fields=2`
- a `slow.fake` file with an accompanying `slow.fake.ini` with `sleepOpenBytes=30000`
- an OME-XML file with an incorrect detector settings reference, currently failing import side
[wrong-detector-settings-mk1.ome.xml.zip](https://github.com/ome/omero-insight/files/12453761/wrong-detector-settings-mk1.ome.xml.zip)

All scenarios were tested individually as well as using combinations and the logs of the server were monitored using the command mentioned in the description.

> regular import of directories (1-10 & >50 to test the lightfileimport)

50 files was the threshold I remembered for the switch to the light file import UI but after verification the current limit is 100. This threshold is configurable using the `importer.xml` https://github.com/ome/omero-insight/blob/40b4f79d484a3e5193ca7c6b5508b1950ff3fb97/src/config/importer.xml#L41 and this value was lowered to 20 for the sake of my testing.

Confirmed that in all scenarios above, all the `HandleTie` servants are appropriately closed when the import completes and no leftover servant ends up being unregistered when closing OMERO.insight and terminating the session.
The only exception to this rule was the testing of the early closure of the importer/insight client after uploading a sample file with a long-processing time (`slow.fake`).  In that case, the servant is not cleaned until the end of the server-side import steps as expected.

A few inline comments/clarifications related to code review. Possibly the biggest impact outside the scope of the original fix is that the `View` button and the checkbox in the standard importer have disappeared.

With OMERO.insight 5.8.1,
Screenshot 2023-08-28 at 11 53 04
Screenshot 2023-08-28 at 11 53 35


With this PR
Screenshot 2023-08-28 at 11 54 08
Screenshot 2023-08-28 at 11 54 21


For import failures, the cross icon is still visible and the `Failed` button gives access to the import logs

Screenshot 2023-08-28 at 12 08 24

I will leave @dominikl to perform his own testing and add his comments. In addition to the above, I would like to add a scenario testing the light UI import with a mixed case of files with successful imports and failures to confirm the statuses are as expected. But that can be scheduled for another round of testing especially if additional commits are added.

src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
jburel: that's  a leftover from a previous attempt. I will remove it

src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ImagesImporter.java
jburel: It can, and the underline method could also be changed i.e. no need to pass a boolean.

src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
jburel: This method is invoked before the "setStatus" and follow a similar logic
```
if (image instanceof ImportException)
            result = image;
        if (result instanceof ImportException) {
```
Light import could be cleaned up too

src/main/java/org/openmicroscopy/shoola/env/data/views/calls/ImagesImporter.java
jburel: done

src/main/java/org/openmicroscopy/shoola/agents/fsimporter/util/FileImportComponent.java
jburel: sorted

sbesson (approved):
Retested with the last commits using all scenarios above as well as a directory using the light UI import with a mixture of successes and failures. Everything works as expected. The completion icon/button are displayed in the importer UI and all servants are unregistered server-side.

dominikl (approved):
Looks good. Tested different scenarios: Few images (with detailed info), many images (with just the summary, set limit to 20 like @sbesson mentioned). Plates. Multiple import tabs. Closing Insight after "File upload" step, etc. UI still behaves like expected. And there's one "Unregistered servant" for each "Added servant" in the Blitz log.
👍

PR #404, opened 2023-11-07
adjust shift logic

### LINKED ISSUE #403: Shift logic may not be correct
https://github.com/ome/omero-insight/blob/40566b97d28f9caf5720a55faf15c324a8c6e5de/src/main/java/org/openmicroscopy/shoola/util/roi/model/util/Coord3D.java#L139C1-L149C3

Regarding this code:

/**
	 * Calculate the hashCode for the data, 
	 * The hashcode is generated by bitshifting
	 * z by ZTBITSPLIT bits and adding t. 
	 * @return see above.
	 */
	public int hashCode()
	{
		int value = z<<ZTBITSPLIT+t;
		return value;
	}

=======================

I believe the shift logic

  z<<ZTBITSPLIT+t

does not do what is described in the comments.

Parenthesis are needed:

(z<<ZTBITSPLIT)+t

jburel:
Fix #403

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/util/roi/model/util/Coord3D.java
+++ b/src/main/java/org/openmicroscopy/shoola/util/roi/model/util/Coord3D.java
@@ -2,7 +2,7 @@
  * org.openmicroscopy.shoola.util.roi.model.util.Coord3D 
  *
   *------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dominikl (approved):
Interesting, I would have expected too that `<<` operator takes precedence over `+`, but parenthesis won't harm anyway and make it clearer 👍

Issue #406, opened 2023-11-12
Is return keyword needed?


scuniff:
Is return keyword needed?

https://github.com/ome/omero-insight/blob/40566b97d28f9caf5720a55faf15c324a8c6e5de/src/main/java/org/openmicroscopy/shoola/util/roi/io/InputStrategy.java#L510

			if (IOConstants.ATTRIBUTE_DATATYPE_RECTANGLE2D.equals(dataType))
				new Rectangle2D.Double(
				new Double(xValue), new Double(yValue), new Double(widthValue),
				new Double(heightValue));

should be?

			if (IOConstants.ATTRIBUTE_DATATYPE_RECTANGLE2D.equals(dataType))
				return new Rectangle2D.Double(
				new Double(xValue), new Double(yValue), new Double(widthValue),
				new Double(heightValue));

jburel:
https://github.com/ome/omero-insight/pull/408

PR #408, opened 2023-11-14
add missing return

jburel:
see #406


dominikl (approved):
Thanks. Makes sense 👍

Issue #421, opened 2024-01-02
Should PlanePoint be changed to PlanePoint2D?


scuniff:
https://github.com/ome/omero-insight/blob/4e8d0754713b0a3c74908b4ebaa9ebae4e378a14/src/main/java/org/openmicroscopy/shoola/util/math/geom2D/PlanePoint2D.java#L232C5-L237C6

Referring to this code snippet:

    public boolean equals(Object o)
    {
    	if (o == null || (!(o instanceof PlanePoint))) return false;
    	 PlanePoint other = (PlanePoint) o;
         return (x1 == other.x1 && x2 == other.x2);  
    }

Should the 3 references to **PlanePoint** actually be **PlanePoint2D**?

PR #422, opened 2024-01-03
use correct type in equals

jburel:
Fix issue #421


dominikl (approved):
👍 Looks good, thanks.

Issue #435, opened 2024-03-18
InvalidPathException: Illegal character [?] in path


dominikl:
```
java.lang.Exception: Abnormal termination due to an uncaught exception.
java.nio.file.InvalidPathException: Illegal character [?] in path at index 2: \\?\U:\RG Affolter...
	at sun.nio.fs.WindowsPathParser.nextSlash(WindowsPathParser.java:212)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:111)
	at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
	at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
	at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
	at java.io.File.toPath(File.java:2236)
	at org.apache.commons.io.FileUtils.isSymlink(FileUtils.java:3107)
	at org.apache.commons.io.FileUtils.sizeOfDirectory0(FileUtils.java:2618)
	at org.apache.commons.io.FileUtils.sizeOf0(FileUtils.java:2641)
	at org.apache.commons.io.FileUtils.sizeOfDirectory0(FileUtils.java:2619)
	at org.apache.commons.io.FileUtils.sizeOfDirectory(FileUtils.java:2599)
	at org.openmicroscopy.shoola.env.data.model.FileObject.getLength(FileObject.java:464)
...
```

See [QA 41496](https://www.openmicroscopy.org/qa2/qa/feedback/41496/), [QA 41497](https://www.openmicroscopy.org/qa2/qa/feedback/41497/). I wonder if that'll probably need an update of the Apache Commons lib...

dominikl:
I just can't replicate the problem. Someone has to provide more information on how they got to this weird `\\?...` path.

jburel:
This probably depends on the Java version/flavour used

dominikl:
Thanks @pwalczysko for the hint. On the MyDesktop/Horizon VM thing I indeed have a similar network drive with a similar weird path. But importing from there works fine, doesn't throw the exception :-/

dominikl:
Another one [QA 41633](https://www.openmicroscopy.org/qa2/qa/feedback/41633/)

jburel:
Space in the path does not seem to be correctly handled. 
See https://www.openmicroscopy.org/qa2/qa/feedback/41640/

dominikl:
[...]

PR #268, opened 2022-01-05
use toURI method to handle

jburel:
This should fix issue with file path.
see https://www.openmicroscopy.org/qa2/qa/feedback/30780/ and https://github.com/ome/omero-insight/issues/237
cc @dominikl

PR #436, opened 2024-03-28
Update apache common libs

dominikl:
I have no idea where the apache commons io dependency is defined. But lets see if this fixes https://github.com/ome/omero-insight/issues/435

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -36,7 +36,8 @@ dependencies {
         exclude group: "com.sun.jna", module: "jna"
     }
     implementation("org.apache.poi:poi:4.0.1")
-    implementation("org.apache.commons:commons-collections4:4.3")
+    implementation("org.apache.commons:commons-collections4:4.4")
+    implementation("commons-io:commons-io:2.15.1")
     implementation("org.apache.httpcomponents:httpmime:4.5.7")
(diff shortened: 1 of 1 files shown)


jburel:
``commons-io`` is injected in omero-romio
There was an attempt to remove it see https://github.com/ome/omero-romio/pull/14
I don't think we should add it
I think we should try to use ``nio Paths`` or ``Files``. See for example https://github.com/ome/omero-insight/pull/268 usage of ``toURI`` to solve a similar problem

jburel:
Actually newer version is injected in omero-server

dominikl:
Good idea. I'll try to remove it 👍

dominikl:
Actually, not a good idea. It's everywhere. I've tried got rid of it in Insight, but then failed at https://github.com/ome/omero-insight/blob/558b56cf66c50c895fd8d5643dc377afd2a4ddee/src/main/java/org/openmicroscopy/shoola/env/data/OmeroImageServiceImpl.java#L263 with `error: cannot access DirectoryWalker, icContainers = ic.getContainers();` I.e. it's heavily used in omero-blitz as well. You mentioned omero-romio... I guess it's in many many more. Which probably also means updating it to a new version is a pain.

dominikl:
--exclude temporary exclude due to merge conflict

PR #442, opened 2024-07-18
Remove duplicate app_name parameter

dominikl:
The `app_name` parameter was included twice. For some reason that caused the failure to send to the new QA.

~~Not ready to merge yet, I want to still check why Insight ignored the error and gave user the wrong impression that the feedback was sent.~~


dominikl:
Ready to merge. Insight actually handles is correctly, it gets a 200 - OK from QA even it fails with the duplicate app_name parameter, needs to be fixed on the other side.

dominikl:
Ready for the next release 👍

pwalczysko:
Works fine with the new QA.

pwalczysko (approved):

PR #443, opened 2024-07-19
Perceptibly uniform luts for OMERO.insight

Tom-TBT:
This PR adds the preview of LUTs introduced in https://github.com/ome/openmicroscopy/pull/6398 to OMERO.insight


jburel:
@dominikl The call is done in the web gateway so nothing more to do in insight

Tom-TBT:
I am closing this PR. If the dynamic LUT preview is implemented in OMERO.insight, it would be done differently than this.

dominikl:
No, I think this PR is correct @Tom-TBT ! Maybe misunderstood @jburel 's comment (that was rather meant for me).

Tom-TBT:
Sorry I maybe went a bit too fast. 

This PR adds new LUTs to the static png file. If the LUTs I added with https://github.com/ome/openmicroscopy/pull/6398 are also included, then this PR still makes sense.

jburel:
This PR updates the LUT to match what has already been introduced in the webclient.
Ideally we could dynamically create the LUT png but considering the cost/benefit, this is not something I will recommend.
Happy to have it in cc @dominikl

PR #444, opened 2024-08-21
fix endpoint

jburel:
This PR fixes the URL when submitting to QA. The bug was introduced in https://github.com/ome/omero-insight/pull/327
The problem was noticed by @pwalczysko when testing the new QA
This prevents the user from submitting a file when an error happens
This might have an impact for the GS QA

### DIFF
--- a/src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
+++ b/src/main/java/org/openmicroscopy/shoola/env/data/DataServicesFactory.java
@@ -558,7 +558,7 @@ public class DataServicesFactory
                 msg.println("Using URL defined server-side for feedback: " + val);
                 registry.getLogger().debug(this, msg);
                 container.getRegistry().bind(LookupNames.TOKEN_URL, val + "/qa/initial/");
-                container.getRegistry().bind(LookupNames.PROCESSING_URL, val + "/qa/uploadProcessing/");
+                container.getRegistry().bind(LookupNames.PROCESSING_URL, val + "/qa/upload_processing/");
             } else {
                 // needed when switching user
                 resetKeys(Arrays.asList(LookupNames.TOKEN_URL, LookupNames.PROCESSING_URL)); 


sbesson:
🤦 yes, the endpoint is wrong and should be `/qa/upload_processing` as in https://github.com/ome/omero-insight/blob/efbf7e35ce95c195c753051f26f92a075e8b5dd3/src/config/container.xml#L180 or https://github.com/ome/omero-blitz/blob/bf9a738b0417973de01be3b2f7ccf45ee2eacd19/src/main/java/ome/formats/importer/ImportConfig.java#L561

> The problem was noticed by @pwalczysko when testing the new QA

Does this mean the old QA works with the `/qa/uploadProcessing` endpoint and this is why this was not flagged as part of the testing of https://github.com/ome/omero-insight/pull/327?
Or has it always been wrong and the QA endpoint can be used for testing the former (broken) behavior as well as the fix?

jburel:
It was not noticed during the testing, the entry was recorded and the message was positive but the file was not uploaded.

pwalczysko:
> existing QA deployment can be used for testing both the current (broken) behavior as well as the fix?
Yes, the current QA (OME Team's) deployment will not get files submitted from the most recent release insight 5.8.4. It will get files submitted from insight 5.7.2.

sbesson (approved):
Tested using OMERO.insight by uploading a sample file known to be invalid and broken during the server-side processing
[wrong-detector-settings-mk1.ome.xml](https://github.com/user-attachments/files/16692233/wrong-detector-settings-mk1.txt) and sending a feedback using the UI.

Using OMERO.insight 5.8.1 i.e. with the changes from #327 but without this fix, an error message was sent but the file upload failed as described. Note this behavior was the same for a OMERO.server configured with `omero.qa.feedback` unset in the server configuration i.e. using the default value (QA 41925) as well as a recent OMERO.server configured with `omero.qa.feedback` explicitly set to http://qa.openmicroscopy.org.uk (QA 41922).

With this PR included the error message and the file were correctly uploaded to the QA app both when `omero.qa.feedback` was unset in the OMERO configuration (QA 41924) and when `omero.qa.feedback` was explicitly set (QA 41923).

In conclusion this issue affected the file upload when OMERO.insight was connecting against any recent deployment of OMERO.server independently of whether `omero.qa.feedback` was set or not. The change fixes the problem in all cases and should probably be released at earliest (possibly with https://github.com/ome/omero-gateway-java/pull/90 which is another regression)

pwalczysko:
Confirming the fix. The insight built from this branch with ``gradle build`` and ``gradle run`` is successfully submitting files. The insight built the same way from master is not.

pwalczysko (approved):

Issue #447, opened 2024-09-16
Review upload/download of artifacts in GHA


omereleases:
using v4 leads to error

sbesson:
Looking at https://github.com/ome/omero-insight/actions/runs/10884718399, looks like the workflow needs to be updated to match the [breaking changes](https://github.com/actions/upload-artifact?tab=readme-ov-file#breaking-changes) in the actions if you want to move to `{upload,download}-artifact@v4`. In particular artifacts are no longer mutable - see also https://github.com/actions/upload-artifact/blob/main/docs/MIGRATION.md#multiple-uploads-to-the-same-named-artifact

jburel:
exactly
I did not have time to look at that now 
@dominikl could you have a look?

pwalczysko:
Possibly fixed by https://github.com/ome/omero-insight/pull/448 - maybe lets wait till next release to close after this :)

jburel:
I think we can close it

PR #448, opened 2024-09-19
Adjust the upload and download of artifacts v4

pwalczysko:
This is trying to address https://github.com/ome/omero-insight/issues/447

This branch starts from the https://github.com/ome/omero-insight/actions/runs/10884718399 - the commit a3ea82d1e59e2630ca3995ee9776a292f74e12ea -> where the build started to fail because of the same-named artifacts.

The strategy taken:

Upload:
1. add the ``matrix.os`` motif to the artifact name -> this is by itself not enough, as we build 2 artifacts per OS on 2 occassions
2. add a hardcoded string such as ``omero_`` which is taken from the foldername of the ``path`` variable just below the ``name`` of the artifact

Download:
1. adjust the ``file: artifacts/*`` -> ``file: artifacts*/*`` so that the artifacts are found in all folders created during Upload.

I have pushed an xxx-test8 tag to my [private branch show-tagged-artifacts](https://github.com/pwalczysko/omero-insight/tree/show-tagged-artifacts) (because I do not want to push tags to this branch so as not to trigger a real release) and this resulted in [Release with artifacts](https://github.com/pwalczysko/omero-insight/releases/tag/v5.8.5-test8)

cc @jburel

### DIFF
--- a/.github/workflows/gradle.yml
+++ b/.github/workflows/gradle.yml
@@ -78,28 +78,28 @@ jobs:
         if: startsWith(matrix.os, 'ubuntu') && startsWith(github.ref, 'refs/tags') && startsWith(matrix.java, '8')
         uses: actions/upload-artifact@v4
         with:
-            name: artifacts
+            name: artifacts-${{ matrix.os }}-OMERO
             path: build/distributions/OMERO*
             if-no-files-found: error
       - name: Upload jar
         if: startsWith(matrix.os, 'ubuntu') && startsWith(github.ref, 'refs/tags') && startsWith(matrix.java, '8')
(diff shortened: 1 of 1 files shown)


pwalczysko:
I have pushed an xxx-test8 tag to my [private branch show-tagged-artifacts](https://github.com/pwalczysko/omero-insight/tree/show-tagged-artifacts) (because I do not want to push tags to this branch so as not to trigger a real release) and this resulted in [Release with artifacts](https://github.com/pwalczysko/omero-insight/releases/tag/v5.8.5-test8)

jburel:
Tested on my account. Everything works as expected
Thanks @pwalczysko

PR #333, opened 2023-01-06
drop compatibility to test new Jenkins

jburel:
Testing on new CI

PR #461, opened 2025-03-20
Update infra

jburel:
* Update actions
* Fix internal dependency
* Remove Java 1.8 support. This no longer works

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@ plugins {
     id "org.openmicroscopy.publish"
     id "org.openmicroscopy.distribute"
     id "org.openmicroscopy.packager"
-    id "org.openmicroscopy.additional-repositories" version "5.5.1"
+    id "org.openmicroscopy.additional-repositories" version "5.7.2"
 }
 
 group = "org.openmicroscopy"
--- a/build.gradle
+++ b/build.gradle
@@ -15,10 +15,10 @@ repositories {
     mavenCentral()
 }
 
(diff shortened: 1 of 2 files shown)


jburel:
https://merge-ci.openmicroscopy.org/jenkins/job/OMERO-insight-build/341/ is back to green
cc @dominikl

sbesson:
> Remove Java 1.8 support. This no longer works

Is this bumping the build requirement to JDK 11 (as we already do on the omero-* components) or the Java runtime requirement? I know we talked about this several times recently, is there a summary of the state of the current packaging issues (might even be a blog post on the OME website)?

jburel:
The compatibility does not work for insight with CI https://github.com/ome/omero-insight/pull/333
This is a build requirement

build.gradle
```diff
@@ -16,8 +16,8 @@ repositories {
 }
 
 java {
-    sourceCompatibility = JavaVersion.VERSION_1_8
-    targetCompatibility = JavaVersion.VERSION_1_8
+    sourceCompatibility = JavaVersion.VERSION_11
```
sbesson, line 19: Assuming JDK 11 is a build requirement only like the other OMERO components, would setting

```
compileJava {
    options.release = 8
}
``` 

be an option here as in https://github.com/ome/omero-model/blob/560fde3c5f344ee4ba5d25c3201b71f3e12adbd0/build.gradle#L32-L34

